PR #46, opened 2026-07-08
Busses: Ableton-style return channels with per-track sends

alexanderwanyoike:
## What

- **Arbitrary return buses**: '+ Bus' in the mixer adds A/B/C... Return channels — mixer-only strips with channel EQ, fader, balance pan, meter, mute, remove control
- **Send knobs on every track strip**: post-fader, colored by bus, lettered; buses have no send knobs so the routing graph can't feed back (Ableton's default behavior)
- **Insert chains on buses**: effects-only device panel (built-ins + CLAP/VST3), selected like any channel
- **Returns use a stereo balance law** — equal-power panning would tax every centered return 3 dB since the send mix is already panned
- **Idle parity**: sends feed buses while the transport is stopped, so auditioned notes reach the reverb return
- **Persistence/undo/bounce**: sends + buses in the project file (legacy files load clean), snapshot replay for undo/redo, plugin reloads on buses, and master-mode bounce/export routes through the buses identically to live playback

## Verified live (Xvfb sandbox)
- Added returns, cranked a Bass send, dropped Reverb on the A Return, watched the return meter light during playback
- Undo chain walks back reverb → send → bus creation correctly

## Gate
21 suites, 0 failed — new tests: send routing math (dry + unity send doubles), bus gain/mute, bus effect chains, remove-bus send cleanup, balance pan law, bounce send parity, project roundtrip + pre-bus schema compat. Clippy -D warnings clean.

## Deferred (noted for later)
- Send-amount automation (needs a new AutomationTarget variant)
- Bus renaming from the strip
- Knob drags create one undo step per motion event — pre-existing behavior across all knobs, worth coalescing separately


alexanderwanyoike:
Extended per discussion: **returns and master are now automatable from the arrangement**, plus deletion everywhere.

- Returns + Master render at the bottom of the arrangement as clipless channel lanes (Ableton-style): slim header with select/expand/delete (returns only — master is not deletable), automation strips identical to track lanes
- Engine evaluates bus and master lanes at block rate: bus gain/pan rides, master gain rides (classic end-of-song fades)
- **Send automation**: new `Send` target — tracks automate their send amount into any bus via the picker ("Send: A Return"); verified live with a 0→1 ramp opening the reverb return mid-song while the mixer knob stayed at zero
- Tracks deletable from their mixer strip, returns from both views
- Bug fix caught by the new tests: **volume automation lanes played at half their displayed level** (UI normalizes gain over 0..2, engine applied the raw value). Now denormalized correctly — existing volume lanes will play louder, at the level the lane always displayed

New engine tests: bus gain lane rides the return, send lane opens a closed send, master gain lane shapes the mix. Gate: 21 suites, 0 failed, clippy clean.

PR #57, opened 2026-07-13
Browser Card 07: audition local media

alexanderwanyoike:
Implements Browser Card 07 as the next chained feature PR.\n\n- replaces the hidden preview path with a dedicated post-master Audition Bus\n- adds click-safe start, replacement, and stop fades plus independent remembered gain\n- keeps Browser audition independent from transport and offline render/capture paths\n- adds the flat waveform, Enabled, Play/Stop, state, and keyboard behavior\n- stops audition on project clear and reconciles decode/completion state\n\nVerification:\n- cargo test -p vibez-engine (128 passed)\n- cargo test -p vibez-ui (142 passed)\n- cargo clippy -p vibez-engine --lib -- -D warnings\n- cargo clippy -p vibez-ui --lib -- -D warnings\n- cargo fmt --all -- --check\n- git diff --check\n- native Xvfb smoke: transport and audition coexist; Escape stops audition only; disabled selection loads waveform; explicit Play still works; gain persists\n\nNative evidence SHA-256:\n- transport + audition: 3aeedafd4894013c9ae31d21d12aea741d85d86739c4276c256ab5a418c0e4c5\n- Escape stop with transport continuing: 544158ac7f45509f995d9508fc7bcdf4fafffc5b3ca3cf7536df7b83e70ad17a\n- disabled selection waveform: 9a2b02bb12e37582e0eb95a07b84fff96b19ae8c59440b3c23e1c9bdd6e52d5a\n- disabled explicit Play: 0579f5d7d80420529d138e9e350de1a2c2381b6b0044f0c8acfef569879e23e5\n- remembered -6 dB gain: 8a6aa183cc4949b6e5388ef761f39e9957e2d2dc209c15138f218b69d75cc052

PR #58, opened 2026-07-13
Browser Card 08: WARP, sync, and loop audition

alexanderwanyoike:
## Demo

Compare a selected Local sample in RAW or pitch-preserving WARP mode, confirm source BPM before WARP, quantize audition start to Beat/Bar only while transport runs, and loop through click-safe boundaries.

## Scope

- adds RAW/WARP mode and explicit confirmed source-BPM input
- adds Off/Beat/Bar audition sync with queued/playing state
- adds persistent Loop defaulting off and click-safe boundary crossfades
- exposes the active audition mode as an explicit import input for Card 09
- keeps Browser styling flat and preserves the responsive Places + Results shell

## Verification

- cargo test -p vibez-engine (133 passed)
- cargo test -p vibez-dsp (64 passed)
- cargo test -p vibez-ui (143 passed)
- cargo clippy -p vibez-engine -p vibez-dsp -p vibez-ui --lib -- -D warnings
- cargo fmt --all -- --check
- git diff --check
- native Xvfb smoke at 300, 410, and 650 px Browser widths; RAW, blocked WARP, stopped transport, BAR queue/start, and loop-on states captured

## Non-goals

Automatic loop/one-shot classification, REPITCH, range-based Remote streaming, and carrying the chosen mode into import (Card 09).

Stacked on #57.

### DIFF
--- a/crates/vibez-dsp/src/time_stretch.rs
+++ b/crates/vibez-dsp/src/time_stretch.rs
@@ -32,14 +32,6 @@ const WSOLA_SEARCH: isize = 256;
 
 // Ratio router.
 const RATIO_PASSTHROUGH_EPS: f64 = 0.005;
-/// Stretch ratios this close to unity use plain resampling instead of
-/// a phase vocoder: the resulting pitch shift is under one semitone
-/// (inaudible on drums, barely audible on melodic material) and the
-/// time domain stays bit-exact, so transients cannot smear. This is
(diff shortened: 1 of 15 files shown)


alexanderwanyoike:
Consolidated into #65 (browser foundation v1, includes this card plus the consensus-review fixes).

PR #68, opened 2026-07-17
refactor(ui): separate project tracks from timeline content

alexanderwanyoike:
## Summary

- represent project-wide channels once in `ProjectTracksState`
- store Arrange clips, note clips, and automation separately in `ArrangementTimeline`, keyed by `TrackId`
- retain independent copy-on-write stores in undo snapshots so timeline edits do not clone devices, routing, or mixer state
- preserve the existing V1 project document shape while hydrating/projecting the new in-memory boundary
- document the ownership boundary in `docs/ARCHITECTURE.md`

## Verification

- legacy project semantic round-trip regression test
- `git diff --check`
- native dogfood: track create/delete, MIDI clip creation and note edit, mixer mute, Save As, New Project, reopen
- largest touched Rust source: 962 lines

## Scope

This delivers Perform Phase 0 / Card 01 only. It does not begin Card 02, Sections, Perform UI, the shared Timeline Editor extraction, or browser-import routing cleanup.

### DIFF
--- a/crates/vibez-ui/src/app/views_arrangement.rs
+++ b/crates/vibez-ui/src/app/views_arrangement.rs
@@ -15,7 +15,7 @@ use vibez_core::id::{ClipId, TrackId};
 
 use crate::icons;
 use crate::message::Message;
-use crate::state::{ArrangementSelection, ContextMenuTarget};
+use crate::state::{ArrangementSelection, ContextMenuTarget, TrackTimelineContent};
 use crate::theme as th;
 use crate::widgets::timeline::{ArrangementMinimap, MinimapTrack, RulerWidget, TrackClipCanvas};
 use crate::widgets::track_header::{view_editable_channel_name, view_track_header};
@@ -208,12 +208,13 @@ impl App {
 
         // Track rows: header widgets + clip canvas
         let mut track_rows = column![].spacing(0);
+        let empty_content = TrackTimelineContent::default();
 
         for (track_index, track) in self.state.project_tracks.tracks.iter().enumerate() {
             let content = self
                 .state
                 .arrange_content(track.id)
-                .expect("every Project Track has Arrange Timeline Content");
+                .unwrap_or(&empty_content);
             let selected = self.state.arrangement.selected_track == Some(track.id);
             let track_color = th::track_color(track.color_index);
 
--- a/crates/vibez-ui/src/app/views_arrangement.rs
+++ b/crates/vibez-ui/src/app/views_arrangement.rs
@@ -24,7 +24,7 @@ use super::*;
 
 impl App {
     pub(super) fn view_arrangement(&self) -> Element<'_, Message> {
-        if self.state.arrangement.tracks.is_empty() {
+        if self.state.project_tracks.tracks.is_empty() {
             let browser_drag_active = self.state.browser.drag_source.is_some();
             let empty_beat = match self.state.browser.drag_target {
                 Some(crate::state::BrowserDropTarget::EmptyArrangement { beat }) => Some(beat),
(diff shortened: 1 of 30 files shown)


crates/vibez-ui/src/app/views_arrangement.rs
```diff
@@ -197,7 +209,11 @@ impl App {
         // Track rows: header widgets + clip canvas
         let mut track_rows = column![].spacing(0);
 
-        for (track_index, track) in self.state.arrangement.tracks.iter().enumerate() {
+        for (track_index, track) in self.state.project_tracks.tracks.iter().enumerate() {
+            let content = self
+                .state
+                .arrange_content(track.id)
+                .expect("every Project Track has Arrange Timeline Content");
```
alexanderwanyoike: This `.expect()` is the only place in the PR where a missing timeline entry panics; every sibling consumer (views_mixer.rs:54/78/132, views_arrangement.rs:535, views_detail.rs:127/203) degrades with `.map(...).unwrap_or(&[])`.

I verified the invariant does hold today: every creation site (`AddTrack`/`AddMidiTrack` via `ensure`, load hydration via the unconditional `arrange_content_mut(track_info.id).automation` assignment, bounce, media import) inserts an entry, and `RemoveTrack` cleans both stores. So this is not a live bug.

But it is an unenforced cross-store invariant guarded by a panic in the render path, and it is exactly the pattern Perform Sections will break by design: Cards 07/10 make absent per-track content an intentional state (missing content = silence). When the Section view is written by adapting this loop, this line becomes an app crash. Please either degrade gracefully like the sibling call sites (render empty content), or keep a `debug_assert!` for the invariant and fall back to `TrackTimelineContent::default()` in release.

alexanderwanyoike (commented):
## Card 01 review (reviewer: Claude, loop session)

Reviewed against `perform/cards/01-separate-project-tracks-from-timeline-content.md` and the delivery contract in `engineering/refactoring-guide.md`. Overall: this is a clean implementation of the seam, and I could not find an Arrange behavior change.

### Independently verified (not just trusting the card evidence)
- `cargo clippy --workspace -j 4 -- -D warnings`: passed on the PR head.
- `cargo test --workspace -j 4`: 439 passed, 0 failed.
- Largest touched Rust file is 962 lines (`domains/arrangement/mod.rs`); the `ops.rs` megafile shrank from 1374 lines, which is a bonus.
- `docs/ARCHITECTURE.md` gained a real ownership-boundary section, not a token edit.
- Ownership: `ProjectTracksState` (tracks/master/buses/channel state) vs `Arc<ArrangementTimeline>` keyed by `TrackId` holding clips/note clips/automation only; `timeline_edits_clone_only_timeline_content` proves pointer identity of the untouched store.
- Save projection iterates the ordered track list (deterministic clip order), pulls automation from the timeline store (`track_info_from_ui`), and the legacy shape round-trips semantically (`existing_project_document_roundtrips_semantically_unchanged`).
- `RemoveTrack` cleans both stores plus selection; undo snapshots pair the two `Arc`s consistently.
- Scope is bounded: no Sections, no Perform UI, no Card 24 import-landing changes smuggled in.

### Findings for sol to address
1. **(fix requested)** `views_arrangement.rs:216`: render-path `.expect()` on an unenforced cross-store invariant; inline comment has the details and two acceptable fixes.
2. **(minor, non-blocking)** `pub type UiTrack = ProjectTrack;` is a fine transition alias, but plan its retirement (Card 05 moves most remaining callers anyway) so the old name does not keep the old ownership model alive.
3. **(observation, no change needed)** Regular tracks always have timeline entries; master/buses get them lazily (on automation load or first edit). All consumers handle absence, but consider one doc comment on `ArrangementTimeline` stating which invariant applies to which channel kind, since the Section store will copy this pattern.

Not approving or merging: phase gates are Alex's. Once finding 1 is fixed and pushed, I will re-review the delta.

crates/vibez-ui/src/app/views_arrangement.rs
alexanderwanyoike: Fixed in f80c449. The arrangement view now falls back to empty TrackTimelineContent when the per-track entry is absent, so the lane renders as silence instead of panicking. Verified with cargo fmt --all -- --check, cargo clippy -p vibez-ui -j 4 -- -D warnings, and cargo test -p vibez-ui -j 4 (212 passed).

alexanderwanyoike: Resolved by `f80c449`: the `.expect()` is now a graceful `unwrap_or(&empty_content)` with the default hoisted outside the loop, matching the sibling call sites. Verified on the new head: workspace clippy clean, vibez-ui tests 212 passed / 0 failed. No further findings; this thread can be closed.

alexanderwanyoike (commented):
## Re-review of f80c449 (reviewer: Claude, loop session)

The single requested fix is applied exactly as asked: `views_arrangement.rs` now falls back to a hoisted `TrackTimelineContent::default()` instead of panicking, consistent with every other content lookup. Delta is 1 file, +3/-2, nothing else changed.

Verified on the new head: `cargo clippy --workspace -- -D warnings` clean; `cargo test -p vibez-ui`: 212 passed, 0 failed (full workspace suite already passed on the parent commit and the delta is view-only).

From my side PR #68 now satisfies Card 01 and the delivery contract with no open findings. Merge and the Phase 0 gate remain with Alex.

PR #69, opened 2026-07-17
fix(ui): group continuous edits into one undo step

alexanderwanyoike:
## Summary

- assign a runtime gesture id to continuous EQ, knob, fader, and timeline clip drag/resize edits
- retain only the first pre-edit snapshot for all incremental messages in one pointer gesture
- keep discrete edits and separate pointer gestures as independent undo steps

## Confirmed cause

The affected widgets emit a project-edit message on every cursor-move event, and the router previously pushed an undo snapshot for every emitted message. A single EQ or clip drag therefore produced several nearly identical history entries. One Undo reverted only the last few pixels, while repeated resize entries hid an earlier automation edit. This cadence already exists on `dev`; Card 01 exposed the pre-existing history granularity defect rather than introducing split-store corruption.

## Regression coverage

- `one_eq_drag_is_one_undo_step`
- `clip_resize_does_not_hide_the_preceding_automation_undo_step`
- existing fader drag tests updated to assert through the gesture wrapper

## Verification

- release build launched for native retest

Stacked on #68; base is `refactor/separate-project-tracks-timeline-content`. Card 02 is untouched.

### DIFF
--- a/crates/vibez-ui/src/widgets/fader.rs
+++ b/crates/vibez-ui/src/widgets/fader.rs
@@ -3,6 +3,7 @@ use iced::widget::canvas;
 use iced::{Color, Rectangle, Renderer, Theme};
 
 use crate::message::Message;
+use crate::state::UndoGestureId;
 use crate::theme;
 use crate::widgets::drag::ValueDrag;
 use vibez_core::id::TrackId;
@@ -30,6 +31,7 @@ impl HorizontalFaderWidget {
 #[derive(Debug, Default)]
 pub struct HorizontalFaderState {
     drag: ValueDrag,
+    undo_gesture: Option<UndoGestureId>,
 }
 
 impl canvas::Program<Message> for HorizontalFaderWidget {
@@ -122,11 +124,13 @@ impl canvas::Program<Message> for HorizontalFaderWidget {
         match event {
             canvas::Event::Mouse(iced::mouse::Event::ButtonPressed(iced::mouse::Button::Left)) => {
                 if state.drag.grab(cursor, bounds, self.value) {
+                    state.undo_gesture = Some(UndoGestureId::new());
                     return (canvas::event::Status::Captured, None);
                 }
             }
             canvas::Event::Mouse(iced::mouse::Event::ButtonReleased(iced::mouse::Button::Left)) => {
                 if state.drag.release() {
+                    state.undo_gesture = None;
                     return (canvas::event::Status::Captured, None);
                 }
             }
@@ -135,7 +139,10 @@ impl canvas::Program<Message> for HorizontalFaderWidget {
                 if let Some(gain) = state.drag.drag_to(cursor, 2.0 / track_w, 0.0, 0.0..=2.0) {
                     return (
                         canvas::event::Status::Captured,
-                        Some(Message::set_track_gain(self.track_id, gain)),
+                        Some(
+                            Message::set_track_gain(self.track_id, gain)
+                                .in_undo_gesture(state.undo_gesture.unwrap()),
+                        ),
                     );
                 }
             }
@@ -171,6 +178,7 @@ mod tests {
 
     fn gain_of(message: Option<Message>) -> Option<f32> {
         match message {
+            Some(Message::UndoGesture { edit, .. }) => gain_of(Some(*edit)),
             Some(Message::Arrangement(ArrangementMsg::SetTrackGain(_, gain))) => Some(gain),
             _ => None,
         }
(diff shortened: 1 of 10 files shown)


crates/vibez-ui/src/widgets/fader.rs
```diff
@@ -135,7 +139,10 @@ impl canvas::Program<Message> for HorizontalFaderWidget {
                 if let Some(gain) = state.drag.drag_to(cursor, 2.0 / track_w, 0.0, 0.0..=2.0) {
                     return (
                         canvas::event::Status::Captured,
-                        Some(Message::set_track_gain(self.track_id, gain)),
+                        Some(
+                            Message::set_track_gain(self.track_id, gain)
+                                .in_undo_gesture(state.undo_gesture.unwrap()),
```
alexanderwanyoike: Minor hardening, applies to all five `state.undo_gesture.unwrap()` sites (fader.rs:144/381, knob.rs:219, effect_knob.rs:397, eq_display.rs:424): the invariant drag-active implies gesture-set is maintained locally today (set on `grab`, cleared on `release`), but these are event-path unwraps in the same class as the render-path `.expect()` from PR #68. `*state.undo_gesture.get_or_insert_with(UndoGestureId::new)` self-heals to a fresh gesture instead of crashing mid-performance if iced ever delivers a stray CursorMoved after state loss. Non-blocking, but cheap insurance in a live-music app.

alexanderwanyoike (commented):
## PR 69 review (reviewer: Claude, loop session)

Not a Perform card, but reviewed on Alex's request as part of the same flow. The mechanism is correct and I could not break it on paper:

- **Gesture lifecycle verified per widget**: fresh `UndoGestureId` at drag-start (`grab`/band-press/`ClipDragAction` construction), wrapped messages during moves, cleared on release. Two separate drags therefore get distinct ids and remain distinct undo steps; a cross-track clip move mid-drag stays inside its drag's single step.
- **History semantics check out**: the first message of any gesture always pushes (and clears redo); only continuation messages early-return, so the redo-clear can never be skipped. Discrete edits, `pop_undo`, `pop_redo`, and `clear` all reset `last_gesture`, so interleaved edits and undo-mid-drag break the group correctly.
- **Router unwrap is clean**: the wrapper is stripped before dirty-classification, so `marks_dirty` runs on the real edit; the `unreachable!` arm is genuinely unreachable.
- **Verified on the head**: workspace clippy clean; `cargo test -p vibez-ui` 214 passed / 0 failed (both new undo tests green).

### Findings
1. **(non-blocking suggestion)** Replace the five `state.undo_gesture.unwrap()` calls with `get_or_insert_with(UndoGestureId::new)`; inline comment on fader.rs:144 has the reasoning.
2. **(test gap, suggested)** The two new tests cover grouping and discrete-edit interleave, but not the most important regression for the id lifecycle: **two separate drags of the same control produce two undo steps**. That is the case that silently breaks if an id is ever hoisted to widget construction or made static. One small test locks it in.
3. **(process note)** Stacked on the Card 01 branch, so merge order is #68 then this; GitHub will retarget it to `dev` automatically when #68 merges. Base choice is correct given it touches `push_undo_snapshot`, which #68 reshaped.

Also worth saying: grouping by gesture id at the history layer, rather than debouncing in widgets or diffing snapshots, is the right design; it will survive Card 15's transaction work cleanly since Section Record sessions can reuse the same collapse-by-key idea.

crates/vibez-ui/src/widgets/fader.rs
alexanderwanyoike: Addressed in 592fb2b: all five event-path unwraps now self-heal with get_or_insert_with(UndoGestureId::new). Added regression coverage at both seams: separate fader drags mint distinct gesture ids, and distinct drag ids create two undo history steps. cargo fmt, strict workspace Clippy, and all 216 vibez-ui tests pass.

alexanderwanyoike: Resolved by `592fb2b`: all five sites now use `get_or_insert_with(UndoGestureId::new)`, and the lifecycle is locked by two new tests, including the widget-level `separate_fader_drags_get_distinct_undo_gestures` driving real press/drag/release event sequences, which is stronger than the history-level test I asked for. Verified on the new head: workspace clippy clean, vibez-ui 216 passed / 0 failed. Thread can be closed.

alexanderwanyoike (commented):
## Re-review of 592fb2b (reviewer: Claude, loop session)

Both findings from my first pass are addressed:

1. The five `unwrap()` sites are now self-healing `get_or_insert_with(UndoGestureId::new)` calls.
2. The id-lifecycle regression is covered twice: `separate_drags_are_separate_undo_steps` at the history layer and `separate_fader_drags_get_distinct_undo_gestures` at the widget layer with real event sequences.

Verified on the head: workspace clippy clean; `cargo test -p vibez-ui` 216 passed / 0 failed. No open findings from me on either #68 or #69. Merge order remains #68 then #69; both gates are Alex's.

PR #72, opened 2026-07-17
Card 03: map computer keys to pad gestures

alexanderwanyoike:
## Summary

- add controller-independent, timestamped Pad Gestures with physical Pad Position, press/release state, optional velocity, and source identity
- route physical computer-key events directly from iced input, pair releases, suppress repeat, and ignore widget-captured typing
- persist the configurable 4x4 key mapping in global UI settings and add a dedicated Perform settings pane
- mirror held gestures across the Pad Surface in Sections, Track Mutes, and Instrument without adding musical behavior

## Verification

- focused Perform, keyboard focus-suppression, global-settings, and project-byte-equivalence tests
- `git diff --check`

Native dogfood covered the default `1234` / `QWER` / `ASDF` / `ZXCV` layout, rebinding R1 · C1 to `Y`, restart persistence, text-field suppression, visible held feedback in all three Perform Modes, and Arrange Ctrl+T / Space / Ctrl+Z regression checks. Local evidence is under `target/card03-dogfood/`, including `card03-pad-mapping.mp4`.

## Scope

Card 03 only. This PR assigns no musical meaning to Pad Gestures and does not begin Card 04, MIDI/VDC input, Instrument velocity, Sections, or launching.

### DIFF
--- a/crates/vibez-ui/src/app/keyboard.rs
+++ b/crates/vibez-ui/src/app/keyboard.rs
@@ -12,6 +12,149 @@ use crate::domains::view::ViewMsg;
 
 use crate::message::Message;
 
+pub(crate) fn keyboard_input_message(
+    event: iced::keyboard::Event,
+    status: iced::event::Status,
+) -> Option<Message> {
+    let should_forward = match &event {
+        iced::keyboard::Event::KeyPressed { .. } => status == iced::event::Status::Ignored,
+        // A release must clear a press that began before focus moved into a
+        // text field. Unpaired releases are harmless in the adapter.
+        iced::keyboard::Event::KeyReleased { .. } => true,
+        iced::keyboard::Event::ModifiersChanged(_) => false,
+    };
+    should_forward.then(|| Message::KeyboardInput {
+        event,
+        occurred_at: std::time::Instant::now(),
+    })
+}
+
+fn computer_key_from_physical(
+    physical: iced::keyboard::key::Physical,
+) -> Option<crate::domains::perform::ComputerKey> {
+    use crate::domains::perform::ComputerKey;
+    use iced::keyboard::key::Code;
+
+    let iced::keyboard::key::Physical::Code(code) = physical else {
+        return None;
+    };
+    Some(match code {
+        Code::Digit0 => ComputerKey::Digit0,
+        Code::Digit1 => ComputerKey::Digit1,
+        Code::Digit2 => ComputerKey::Digit2,
+        Code::Digit3 => ComputerKey::Digit3,
+        Code::Digit4 => ComputerKey::Digit4,
+        Code::Digit5 => ComputerKey::Digit5,
+        Code::Digit6 => ComputerKey::Digit6,
+        Code::Digit7 => ComputerKey::Digit7,
+        Code::Digit8 => ComputerKey::Digit8,
+        Code::Digit9 => ComputerKey::Digit9,
+        Code::KeyA => ComputerKey::A,
+        Code::KeyB => ComputerKey::B,
+        Code::KeyC => ComputerKey::C,
+        Code::KeyD => ComputerKey::D,
+        Code::KeyE => ComputerKey::E,
+        Code::KeyF => ComputerKey::F,
+        Code::KeyG => ComputerKey::G,
+        Code::KeyH => ComputerKey::H,
+        Code::KeyI => ComputerKey::I,
+        Code::KeyJ => ComputerKey::J,
+        Code::KeyK => ComputerKey::K,
+        Code::KeyL => ComputerKey::L,
+        Code::KeyM => ComputerKey::M,
+        Code::KeyN => ComputerKey::N,
+        Code::KeyO => ComputerKey::O,
+        Code::KeyP => ComputerKey::P,
+        Code::KeyQ => ComputerKey::Q,
+        Code::KeyR => ComputerKey::R,
+        Code::KeyS => ComputerKey::S,
+        Code::KeyT => ComputerKey::T,
+        Code::KeyU => ComputerKey::U,
+        Code::KeyV => ComputerKey::V,
+        Code::KeyW => ComputerKey::W,
+        Code::KeyX => ComputerKey::X,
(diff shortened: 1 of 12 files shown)


crates/vibez-ui/src/app/keyboard.rs
```diff
@@ -12,6 +12,149 @@ use crate::domains::view::ViewMsg;
...
+        Code::KeyW => ComputerKey::W,
+        Code::KeyX => ComputerKey::X,
+        Code::KeyY => ComputerKey::Y,
+        Code::KeyZ => ComputerKey::Z,
+        _ => return None,
+    })
+}
+
+fn runtime_key_id(key: &iced::keyboard::Key) -> String {
+    format!("{key:?}")
```
alexanderwanyoike, line 83: **Stuck-pad edge case**: press/release pairing is keyed by `runtime_key_id(&key)`, the Debug string of the *logical* key, while press mapping resolves via the *physical* key. The logical key changes with modifiers, so this sequence desyncs:

1. Press `Q` (no modifiers) → press gesture, active entry keyed `Character("q")`.
2. Press and hold Shift, then release `Q` → the release event's logical key is `Character("Q")` → no matching entry → treated as unpaired and ignored.
3. The stale `"q"` entry keeps `is_pad_pressed` true (stuck highlight), and the *next* plain press of `Q` is swallowed by repeat suppression → one missed press gesture. Recovery only happens on that press's release.

Cosmetic today, but Card 04 turns these gestures into mutes and Card 12 into notes, where a swallowed press is a missed note mid-performance. Fix is small: key `active_computer_keys` by the physical `Code` (the same identity used for mapping) instead of the logical key string; `KeyReleased` carries `physical_key` too. That also drops the per-keystroke `format!` allocation.

alexanderwanyoike (commented):
## Card 03 review (reviewer: Claude, loop session)

Reviewed against `perform/cards/03-map-computer-keys-to-pad-gestures.md` and the delivery contract. The adapter architecture is exactly what Decision 0010 asked for; one fix requested.

### Independently verified
- `cargo clippy --workspace -- -D warnings`: clean; `cargo test --workspace`: 640 passed, 0 failed.
- File sizes fine (perform.rs 698, keyboard.rs 458); `docs/ARCHITECTURE.md` updated.
- **Gesture value**: `PadGesture` carries position, press/release, optional velocity, source identity, and an `Instant` stamped in the subscription closure at input-event time; dispatch is synchronous through the router, with no `Message::Tick` dependency. The card's no-render-loop criterion is genuinely met at the UI-architecture level (sub-frame engine scheduling is correctly deferred to the cards that consume gestures musically).
- **Mode-independence**: gestures resolve through the mapping before any mode logic; tested across Sections/Instrument with a rebound key.
- **Focus suppression**: presses forward only when the event status is `Ignored`; releases always forward so a hold that predates focus loss clears cleanly. Unit-tested plus dogfooded against a real text field.
- **Bonus**: routing all keyboard input through `listen_with` also fixed my Card 02 observation; F1-F3 and the other global shortcuts are now suppressed in text fields via the same path.
- **Mapping**: default physical `1234/QWER/ASDF/ZXCV`, swap-on-conflict rebinding (no two pads can share a key), persisted in `ui.json` under a serde default, restored at startup, absent from projects with a byte-equivalence regression test. Escape cancels a pending rebind and closing Settings clears it.
- Scope bounded: gestures have no musical meaning; nothing from Cards 04+ leaked in.

### Finding
1. **(fix requested)** Press/release pairing is keyed by the *logical* key's Debug string while mapping resolves via the *physical* key, so a Shift held across a release desyncs the pair: stuck pad highlight plus one swallowed press. Inline comment on keyboard.rs has the repro and the fix (key the active map by physical `Code`). Cheap now, painful after Card 04 gives presses musical meaning.

Will re-review the delta when the fix lands.

crates/vibez-ui/src/app/keyboard.rs
alexanderwanyoike: Thanks for flagging the press/release invariant. I traced this through the exact iced_winit 0.13 conversion path before changing it.

On native targets, iced constructs `Event::KeyPressed.key` and `Event::KeyReleased.key` from winit `event.key_without_modifiers()` (`iced_winit/src/conversion.rs:195-200`), while the modifier-sensitive logical value is exposed separately as `modified_key` on the press event. This adapter uses `key`, not `modified_key`, so the proposed sequence remains `Character("q")` on both the initial press and the Shift-modified release. I therefore cannot reproduce the stated desync on the Linux/macOS/Windows targets supported and exercised by this PR.

There is also no `physical_key` field on iced 0.13 `KeyReleased`; it contains only `key`, `modifiers`, and `location` (`conversion.rs:262-267`). That means the suggested physical-`Code` pairing is not directly available at the application boundary.

The WASM conversion is different—it currently falls back to `logical_key`—but Vibez does not have a WASM target. The Debug-string allocation is still a reasonable cleanup opportunity, but it is separate from this native correctness finding. I suggest resolving this thread as not reproducible against the shipped stack; I can add an explicit modifier-transition regression test if we want the iced contract documented locally.

alexanderwanyoike: Confirmed against the vendored source, and I withdraw the fix request. In `iced_winit-0.13.0/src/conversion.rs` the same `key` binding (built from `event.key_without_modifiers()` on non-wasm) feeds both the `KeyPressed` and `KeyReleased` arms, so the Shift-held release delivers the same logical value as the press and the desync I described cannot occur on the shipped targets. `iced_core-0.13.2`'s `KeyReleased` indeed carries only `key`/`location`/`modifiers`, so the physical-Code pairing is not expressible at the release boundary anyway. My finding was wrong on this stack; good trace.

Please do add the modifier-transition regression test you offered: the guarantee lives in iced's conversion layer, not in this codebase, and the wasm arm of that same match already behaves differently (`logical_key` fallback). A local test pins the contract so an iced upgrade that changes the conversion fails loudly here instead of silently reintroducing exactly this bug class. With that test, no open findings from me on #72.

PR #73, opened 2026-07-17
Perform: toggle Project Track mutes from pads

alexanderwanyoike:
## Summary

- map Track Mutes pads to stable Project Track slots with independent bank navigation
- route pointer and mapped-key presses through the shared Project Track mute state
- emit engine-effective mute events with absolute transport sample timestamps
- keep Arrange, Mix, Perform, persistence, and the engine aligned

## Verification

- cargo fmt --all -- --check
- CARGO_TARGET_DIR=target/card04-cargo nice -n 10 cargo clippy --workspace -j 4 -- -D warnings
- CARGO_TARGET_DIR=target/card04-cargo nice -n 10 cargo test --workspace -j 4
- CARGO_TARGET_DIR=target/card04-cargo nice -n 10 cargo build --workspace -j 4
- focused Perform, router, bank-stability, shortcut, and engine timestamp tests

## Dogfood

Recorded native dogfood during Arrange playback verified F2 mode selection, mapped-key and pointer toggles, mixer-to-pad and pad-to-mixer consistency, and persistence across mode/workspace switches. Local recording: target/card04-dogfood/card04-track-mutes.mp4.

## Scope

Card 04 only. Starts from merged Card 03 on dev at bf106fc. Sections, quantized changes, Capture, Solo pads, and Card 05 are intentionally excluded. Merge remains the human gate.

### DIFF
--- a/crates/vibez-ui/src/app/actions.rs
+++ b/crates/vibez-ui/src/app/actions.rs
@@ -14,11 +14,15 @@ use crate::state::{ArrangementSelection, AuditionMode, DetailPanelTab, UiEffect}
 use super::*;
 
 fn apply_track_mute_request(
-    project_tracks: &mut crate::state::ProjectTracksState,
+    project_tracks: &mut Arc<crate::state::ProjectTracksState>,
+    history: &mut crate::state::UndoHistory,
+    pre_edit_snapshot: crate::state::ProjectSnapshot,
     request: crate::domains::perform::TrackMuteRequest,
     engine: &mut impl crate::domains::EngineHandle,
 ) -> Option<String> {
-    let track = project_tracks.find_mut(request.track_id)?;
+    project_tracks.find(request.track_id)?;
+    history.push_edit(pre_edit_snapshot, None);
+    let track = Arc::make_mut(project_tracks).find_mut(request.track_id)?;
     track.mute = request.muted;
     let track_name = track.name.clone();
     engine.send(EngineCommand::SetTrackMute(request.track_id, request.muted));
@@ -34,10 +38,13 @@ impl App {
             self.state.status_text = "Perform key mapping saved".into();
         }
         if let Some(request) = action.track_mute_request {
+            let pre_edit_snapshot = self.take_snapshot();
             let changed = {
                 let mut engine = crate::domains::EngineTx(&mut self.cmd_tx);
                 apply_track_mute_request(
-                    Arc::make_mut(&mut self.state.project_tracks),
+                    &mut self.state.project_tracks,
+                    &mut self.state.project.history,
+                    pre_edit_snapshot,
                     request,
                     &mut engine,
                 )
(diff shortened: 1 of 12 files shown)


crates/vibez-ui/src/app/actions.rs
```diff
@@ -13,7 +13,45 @@ use crate::state::{ArrangementSelection, AuditionMode, DetailPanelTab, UiEffect}
...
+    engine.send(EngineCommand::SetTrackMute(request.track_id, request.muted));
+    Some(track_name)
+}
+
 impl App {
+    /// Apply cross-domain effects requested by Perform without giving the
+    /// Perform interaction slice ownership of Project Track state.
+    pub(super) fn apply_perform_action(&mut self, action: crate::domains::perform::PerformAction) {
+        if action.persist_settings {
+            self.persist_ui_settings();
```
alexanderwanyoike, line 37: **Undo inconsistency between the two mute entry points**: `apply_perform_action` calls `mark_project_dirty()` directly, bypassing the router's snapshot path, so a pad mute creates no undo step. The mixer's `ArrangementMsg::SetTrackMute` is not in the `marks_dirty()` exclusion list, so the identical operation from the mixer pushes an undo snapshot. Same shared `ProjectTrack.mute`, two different histories.

Concrete producer-facing consequence: toggle a pad mute, then press Ctrl+Z. Instead of unmuting, undo reverts the *previous* edit, and because `ProjectSnapshot` includes `project_tracks`, that earlier snapshot ALSO silently reverts the mute as collateral. The mute state becomes dependent on unrelated undo traffic.

Either direction is defensible; pick one deliberately:
- **(a) Pad mutes are project edits like mixer mutes** → route the request through a snapshot (e.g. classify via the same dirty/undo path) so both entry points are undoable.
- **(b) Live mutes are performance state, not edits** → then the mixer path should arguably match during Perform, and at minimum the divergence needs a comment plus a decision recorded in the Perform notes, because CONTEXT calls Track Mute "live Perform state" while this card grounds it in persisted project state.

My lean is (a) for V1: this card's mute IS the persisted mixer mute (it lands in `.vzp` via `TrackInfo.mute`), and un-undoable persisted edits are a trap. The Card 17/18 capture work can revisit live-mute semantics when Capture makes the distinction real.

alexanderwanyoike (commented):
## Card 04 review (reviewer: Claude, loop session)

Reviewed against `perform/cards/04-toggle-track-mutes-from-pads.md` and the delivery contract. The ownership architecture is exactly what the card demanded after sol's earlier grill: Perform stores no second mute value, requests resolve against the shared `ProjectTrack.mute`, the router applies them, and the engine emits the authoritative timestamped event. One fix requested.

### Independently verified
- `cargo clippy --workspace -- -D warnings`: clean; `cargo test --workspace`: 646 passed, 0 failed.
- **Engine truth**: `TrackMuteChanged { track_id, muted, effective_at_samples }` is emitted from the command drain with the transport sample position, only when the track exists, and is covered by an engine test asserting the exact sample. This is precisely the event Card 18's Capture needs.
- **Single mute state**: pad visuals and toggle requests both read `track.mute` through `track_for_mute_pad`; the UI also mirrors `TrackMuteChanged`, so mixer-originated mutes update pads and vice versa. The router test proves state and engine command move together.
- **Stable slots**: `track_mute_slots` vacates removed tracks, fills vacancies with new tracks, and never reorders survivors, which satisfies the no-scrambling criterion; bank clamping shrinks when slots shrink.
- **Bank keys**: `[` / `]` route through the suppressed keyboard path, gate on visible + Track Mutes mode, and are tested.
- **Request-once semantics**: repeat-suppressed presses and releases produce no second toggle, tested.
- Scope bounded: no momentary mute, Solo, groups, quantized changes, Sections, or Capture.

### Finding
1. **(fix requested)** Pad mutes bypass the undo snapshot path while identical mixer mutes create undo steps; the inline comment on actions.rs has the Ctrl+Z failure narrative and two resolutions (my lean: make pad mutes undoable like mixer mutes for V1, revisit at Capture time).

### For Alex's gate: this is a feel card
Automated review cannot judge the two criteria that matter most here. Before merging, worth ten minutes hands-on:
- **Toggle latency**: does a pad key mute feel immediate against a playing project, through the full key -> gesture -> router -> command-drain path?
- **Glitch listen**: the engine applies `track.mute` as a per-buffer bool with no ramp, same as the mixer today, so any click on mute/unmute is pre-existing rather than new, but Track Mutes mode invites rhythmic toggling that will expose it much harder than the mixer ever did. If it clicks musically, that becomes a new card (short gain ramp), not a blocker here.

Will re-review the delta when the undo fix lands.

crates/vibez-ui/src/app/actions.rs
alexanderwanyoike: Fixed in 7444122. Pad mute requests now retain a pre-edit ProjectSnapshot and push it through the same UndoHistory path before mutating the shared ProjectTrack, so Ctrl+Z restores the mute itself instead of consuming an unrelated edit. Invalid/stale track requests create neither an undo step nor an engine command. Added regression coverage for both cases; cargo test -p vibez-ui is 239/239 and workspace clippy passes with -D warnings.

alexanderwanyoike: Resolved by `7444122` with option (a): pad mutes now push the pre-edit snapshot through `history.push_edit(snapshot, None)` before mutating, so both mute entry points are equally undoable, each pad toggle is its own discrete undo step (the `None` gesture also correctly breaks any in-flight drag grouping), and the existence check ensures a request against a missing track creates no history entry. Both new tests cover it. Verified on the head: clippy clean, vibez-ui + vibez-engine tests 375 passed / 0 failed. Thread can be closed.

alexanderwanyoike (commented):
## Re-review of 7444122 (reviewer: Claude, loop session)

The undo-consistency fix is applied as recommended: pad mutes and mixer mutes now share identical undo semantics, with pre-edit snapshot capture, discrete per-toggle steps, and a guarded no-op path, all covered by tests. Delta is 1 file, +58/-8, nothing else changed.

Verified: clippy clean; vibez-ui and vibez-engine suites 375 passed / 0 failed.

No open findings on #73. Remaining before merge is Alex's hands-on feel pass: toggle latency against a playing project, and the mute-click listen (pre-existing engine behavior, but Track Mutes mode will expose it far harder than the mixer did; if it clicks musically, that is a new ramp card, not a blocker here).

PR #76, opened 2026-07-18
feat(perform): author MIDI section content

alexanderwanyoike:
## Summary

- add `SectionTimelineEditor` as the second consumer of the shared Timeline Editor boundary
- route shared clip and piano-roll editing to the selected Section while preserving project-wide Project Track selection
- render compact and deliberately expanded multitrack Section timelines with MIDI clip/note authoring
- commit authored Section content into the canonical project document for save/reopen
- keep Section authoring disconnected from the audible Arrange engine source until Section playback lands

## Blockers and scope

Card 05 and Card 06 are satisfied by merged PRs #74 and #75. This branch starts from current `dev` at `60d4760`.

Non-goals: audio Section clips, non-destructive Section length behavior, large-project scrolling, Project Track deletion UX, Section launch/playback, Capture, and any new Arrange editor feature.

## Verification

- `cargo fmt --all -- --check`
- `cargo check --workspace -j4`
- `cargo test --workspace -j4` (UI: 253 passed, 0 failed)
- `cargo clippy --workspace -j4 -- -D warnings`
- `git diff --check`
- every touched Rust file is at or below 1,000 lines
- native dogfood used the exact Card 07 target executable and verified create MIDI clip, draw C4 note, compact/expanded timeline, save/reopen persistence, and unchanged Arrange content

## Manual sign-off

1. Open a project with at least one MIDI Project Track and switch to Perform.
2. Click an empty Section pad, then click the + button on a MIDI row.
3. Click the new clip, switch the piano roll to Draw, and add a few notes.
4. Toggle the timeline expand button in the Section toolbar and return to compact view.
5. Save As, reopen the project, and confirm the Section clip/notes remain while Arrange is unchanged.

Card 08 will be a separate stacked PR based on this branch and will add audio authoring, length-boundary behavior, and scrolling only.

PR #77, opened 2026-07-18
refactor(engine): separate playback source

alexanderwanyoike:
## Summary

- separate time-based clips, note clips, and automation into `PreparedPlaybackSource`
- keep instruments, effects, sends, channel controls, meters, and render scratch on the shared `EngineTrack` channel strip
- make `ArrangementPlaybackSource` the first active adapter while retaining the existing lock-free command and render behavior
- document the fully resident ownership-transfer contract that Card 10 will exercise

## Blockers and scope

Card 01 is complete. This branch starts independently from current `dev` at `60d4760` and has no dependency on Card 07 PR #76.

Non-goals: the Section adapter, source switching, launch behavior, immediate note release, quantized scheduling, transition events, and Capture.

## Real-time contract

A prepared source contains decoded clip `Arc` values and exposes no loader, lock, or I/O API. Its owned pointer is allocated before playback. Card 10 must transfer prepared ownership in the callback and return the superseded owner for disposal outside the real-time thread; it must not drop that owner in the callback.

## Verification

- `cargo test --workspace -j4` (engine: 136 passed; UI: 248 passed)
- `git diff --check`
- touched Rust files remain below 1,000 lines
- branch-exact native dogfood loaded and played the four-track demo Arrange project through 17.56 seconds with transport, playhead, and master metering active, then stopped cleanly on an isolated display

The initial workspace link exhausted the dedicated target disk allocation. Only regenerable Card 09 target artifacts were cleaned; the complete gates then passed with incremental output and debug info disabled.

### DIFF
--- a/crates/vibez-engine/src/engine.rs
+++ b/crates/vibez-engine/src/engine.rs
@@ -13,9 +13,9 @@ use crate::commands::{AuditionSync, EngineCommand};
 use crate::events::EngineEvent;
 use crate::metering;
 use crate::mixer::{
-    any_solo, calculate_total_length, equal_power_pan, EffectSlot, EngineClip, EngineNoteClip,
-    EngineTrack,
+    any_solo, equal_power_pan, EffectSlot, EngineClip, EngineNoteClip, EngineTrack,
 };
+use crate::playback_source::calculate_total_length;
 use crate::transport::Transport;
 
 /// Capacity of the UI spectrum-analyser sample ring: roughly a third
@@ -818,7 +818,12 @@ impl AudioEngine {
         } else {
             0.0
         };
-        let total = calculate_total_length(&self.tracks, samples_per_beat);
+        let total = calculate_total_length(
+            self.tracks
+                .iter()
+                .map(|track| track.playback_source.as_ref()),
+            samples_per_beat,
+        );
         if total > 0 {
             self.transport.set_audio_length(Some(total));
         } else if self.audio.is_none() {
--- a/crates/vibez-engine/src/engine_drain_commands.rs
+++ b/crates/vibez-engine/src/engine_drain_commands.rs
@@ -77,7 +77,7 @@ impl AudioEngine {
                     loop_end,
                 } => {
                     if let Some(track) = self.tracks.iter_mut().find(|t| t.id == track_id) {
-                        track.clips.push(EngineClip {
+                        track.playback_source.clips.push(EngineClip {
                             id: clip_id,
                             audio,
                             position,
@@ -92,7 +92,7 @@ impl AudioEngine {
                 }
                 EngineCommand::RemoveClip(track_id, clip_id) => {
                     if let Some(track) = self.tracks.iter_mut().find(|t| t.id == track_id) {
-                        track.clips.retain(|c| c.id != clip_id);
+                        track.playback_source.clips.retain(|c| c.id != clip_id);
                     }
                     self.recalculate_audio_length();
                 }
@@ -106,7 +106,12 @@ impl AudioEngine {
                     loop_end,
                 } => {
                     if let Some(track) = self.tracks.iter_mut().find(|t| t.id == track_id) {
-                        if let Some(clip) = track.clips.iter_mut().find(|c| c.id == clip_id) {
+                        if let Some(clip) = track
+                            .playback_source
+                            .clips
+                            .iter_mut()
+                            .find(|c| c.id == clip_id)
+                        {
                             clip.audio = audio;
                             clip.duration = duration;
                             clip.source_offset = source_offset;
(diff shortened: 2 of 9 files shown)


alexanderwanyoike (commented):
## Card 09 review (reviewer: Claude, loop session)

Reviewed against `perform/cards/09-deepen-engine-playback-source.md` and the delivery contract. The seam is appropriately minimal per the card's narrowed scope; no fix requested, two observations that become Card 10 requirements.

### Independently verified
- `cargo clippy --workspace -- -D warnings`: clean; `cargo test --workspace`: 657 passed, 0 failed, including the untouched engine behavior suites, which is the honest Arrange-identical evidence (the new identity test compares two tracks that both run the shared path, so the pre-existing suite is what proves nothing changed audibly).
- **Separation is real**: `EngineTrack` keeps channel identity, gain/pan/mute/solo, effects, sends, instrument, meters, and scratch buffers; clips, note clips, and automation moved into `PreparedPlaybackSource`, which exposes no loading or I/O API by construction. `mixer.rs` shrank from 1117 to 957 lines, dropping it back under the limit as a side effect.
- **Minimal, not speculative**: no switching machinery, no quantized scheduling, no Section adapter; `ArrangementPlaybackSource` is the single adapter and the contract comment explicitly defers switch behavior to Card 10, matching the card's out-of-scope list.
- **Callback discipline preserved**: the render path only reads resident `Arc` audio; no allocation or I/O was introduced.
- `docs/ARCHITECTURE.md` updated with the boundary.

### Observations, both aimed at Card 10 (no change requested here)
1. **The `Deref<Target = PreparedPlaybackSource>` bridge should be temporary.** It keeps this diff reviewable by letting every existing `track.clips` site compile, but it also lets code keep treating channel strip and content as one object, which is the exact conflation this seam exists to end. Card 10 should access `playback_source` explicitly at the sites it touches, and the Deref impls should be deleted once call sites migrate; if they survive to Card 12, they will have become load-bearing.
2. **Source switching must transfer the old `Box` out, never drop it in the callback.** Replacing `Box<PreparedPlaybackSource>` in the audio thread would deallocate the outgoing source's Vecs and Arcs on the callback, violating the allocation-free rule from the drop side. The prepared-owner-transfer design hinted in the module comment (swap and ship the old owner back through a queue for off-thread drop) is the right shape; making that explicit is Card 10's first test to write.

From my side #77 satisfies Card 09; merge gate is Alex's, and Card 10 validates this boundary with its second adapter immediately.

alexanderwanyoike:
Review acknowledged—no Card 09 change is needed. I have preserved both observations as Card 10 constraints: remove the temporary `Deref` bridge at touched sites, and transfer the outgoing prepared source off the callback before drop. Card 10 remains unstarted until the human merge gates for #76 and #77 are satisfied.

alexanderwanyoike (commented):
## Revision: observation 1 upgraded to fix requested (reviewer: Claude)

Applying a stricter reusability/architecture bar at Alex's direction, my earlier observation about the `Deref<Target = PreparedPlaybackSource>` bridge is now a fix request on this PR rather than a Card 10 note. Reasoning: the bridge lets every call site keep treating channel strip and content as one object, which is the precise conflation this seam PR exists to end, and transitional bridges without a removal commitment historically outlive their welcome.

Acceptable resolutions, either is fine:
1. Remove the Deref impls in this PR and migrate the call sites to explicit `playback_source` access (the compiler enumerates them; the diff grows but stays mechanical); or
2. Keep the bridge but add `#[deprecated(note = "access playback_source explicitly; removed in Card 10")]`-style enforcement or a `// REMOVE(card-10)` marker plus a named commitment in the Card 10 notes, so the removal is tracked work rather than an intention.

The off-thread-drop requirement for source switching stands as previously written.

alexanderwanyoike:
Addressed the substantive rolling-review request in `e75caba`.

- Removed `Deref<Target = PreparedPlaybackSource>` and `DerefMut` entirely.
- Migrated engine commands, automation, offline rendering, and tests to explicit `playback_source` access, so channel-strip state and timeline-source state can no longer be conflated implicitly.
- Preserved the separate Card 10 constraint that an outgoing prepared source must be transferred off the callback before destruction.

Verification at the pushed head: formatting/diff checks, workspace check, full workspace tests, strict all-target engine Clippy, exact `vibez` build, and exact-binary smoke run all pass. Engine: 136 tests; UI: 248 tests. The local Rust 1.96.1 full-workspace all-targets Clippy command additionally reports two pre-existing `vibez-plugin-host` `unnecessary_cast` warnings unchanged on `origin/dev`; Card 09 code is strict-Clippy clean.

alexanderwanyoike (commented):
## Re-review of e75caba (reviewer: Claude, loop session)

The upgraded fix request is resolved with the stronger of the two options: the `Deref`/`DerefMut` impls are deleted outright and every call site (drain commands, render, tests) now accesses `playback_source` explicitly, so channel strip and timeline content can no longer be conflated by accident. No bridge survives to Card 10.

Verified on the head: workspace clippy clean; full workspace suite 657 passed / 0 failed, so the migration is behavior-preserving.

No open findings on #77. All three open PRs (76, 78, 77) are now clean from my side; the remaining Card 10 requirement (source switch transfers the old Box out for off-thread drop, never dropping in the callback) is recorded on this PR and in my loop state for the Card 10 review.

PR #78, opened 2026-07-18
feat(perform): complete Section authoring (stacked on #76)

alexanderwanyoike:
## Dependency

This PR is deliberately stacked on #76 (`feat/perform-section-midi-authoring`) at commit `bab9b4b`. Do not merge this PR before #76. Once #76 passes the human merge gate, retarget this PR to `dev` without mixing either card scope.

## Demo outcome

Build a complete multitrack Section with MIDI, audio, and shared-editor automation; shorten its playable boundary without destroying hidden content; safely distinguish Section-only removal from project-wide Track deletion; and scroll large Track lists.

## What changed

- Routes Browser audio drops into the selected Section, rejecting MIDI lanes and leaving Arrange untouched.
- Reuses the shared automation editor for Section-local Volume and Pan lanes.
- Treats Section length as a non-destructive visibility/playable boundary and persists out-of-bound content.
- Adds Section-only lane content removal.
- Adds a project Track deletion confirmation listing Arrange and every affected Section; confirmation remains one undoable edit.
- Adds vertical scrolling for large Section Track lists and exposes Browser alongside Perform for authoring.

## Non-goals

- Section launch or playback (Cards 10 and 11)
- Section Record (Card 16)
- Any editor capability unavailable in Arrange

## Verification

- `cargo fmt --all -- --check`
- `cargo check --workspace -j4`
- `cargo clippy --workspace -j4 -- -D warnings`
- `cargo test --workspace -j4` (including 258 `vibez-ui` and 136 `vibez-engine` tests)
- Exact-branch native dogfood on isolated Xvfb `:98`: audio authoring and MIDI rejection; automation editing; shorten/save/reopen/extend identity restoration; affected-location deletion warning; and vertical scroll through an eight-Track Section.

Manual UI sign-off is requested before the human merge gate.

PR #79, opened 2026-07-19
feat(ui): resize the shared detail panel

alexanderwanyoike:
## Outcome

The shared Clip/Devices detail panel is vertically resizable in both Arrange and Perform. Its height persists globally and reclamps on smaller windows so the workspace remains usable.

## Scope

- reuses the shared shell and one pane-splitter visual primitive
- adds begin/resize/end view-domain state and persists only on release
- keeps both Clip and Devices tabs at the chosen height
- no Card 10 playback changes and no unrelated Arrange/Perform styling

## Evidence

- red-first drag lifecycle, geometry/clamping, legacy settings, and round-trip tests
- `cargo test --workspace -j4` (270 UI tests; 136 engine tests)
- exact branch executable `/home/alexander/Code/Apps/vibez-targets/fix-bottom-inspector/debug/vibez` verified by window PID and `/proc`; native Xvfb dogfood resized 280px to 446px and verified persistence after restart
- screenshots: `inspector-before.png`, `inspector-resized-slow.png`, `inspector-reopened.png` in the dedicated target

## Manual demo

1. Open Arrange or Perform and select a Track/clip so the bottom panel is visible.
2. Drag the thin horizontal divider above Clip/Devices up and down.
3. Switch between Clip and Devices and between Arrange and Perform; the same height should remain.
4. Restart Vibez; the chosen height should persist.
5. Shrink the window and confirm neither the workspace nor panel collapses.

### DIFF
--- a/crates/vibez-ui/src/app/mod.rs
+++ b/crates/vibez-ui/src/app/mod.rs
@@ -246,6 +246,7 @@ impl App {
             confirm_project_track_deletion: ui_settings.confirm_project_track_deletion,
             view: crate::state::ViewState {
                 perform_surface_width: ui_settings.perform_surface_width,
+                detail_panel_height: ui_settings.detail_panel_height,
                 ..Default::default()
             },
             browser: crate::state::BrowserState {
--- a/crates/vibez-ui/src/app/project_io.rs
+++ b/crates/vibez-ui/src/app/project_io.rs
@@ -230,6 +230,7 @@ impl App {
             sample_browser_open: self.state.browser.open,
             sample_browser_width: self.state.browser.dock_width,
             perform_surface_width: self.state.view.perform_surface_width,
+            detail_panel_height: self.state.view.detail_panel_height,
             audition_enabled: self.state.browser.audition_enabled,
             audition_gain: self.state.browser.audition_gain,
             audition_loop: self.state.browser.audition_loop,
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -254,6 +254,30 @@ impl App {
                         self.persist_ui_settings();
                     }
                 }
+                let detail_panel_resize = match &msg {
+                    ViewMsg::CursorMoved(_, y) if self.state.view.detail_panel_resize_active => {
+                        Some(ViewMsg::ResizeDetailPanel(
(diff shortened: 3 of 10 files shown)


alexanderwanyoike (commented):
## PR 79 review (reviewer: Claude, loop session)

Not a numbered card, but reviewed under the same contract and strict bar. No findings.

- **Reuse done right**: instead of copying the horizontal splitter's styling into a vertical twin, the shared `pane_splitter_style` was extracted and both splitters consume it; the vertical variant mirrors the horizontal signature exactly. This is the pattern the strictness policy exists to encourage.
- The resize gesture reuses the established begin/resize/end view-domain shape with persist-on-release only, same as the Perform surface splitter, with the matching one-commit-per-drag test.
- Clamping sits at the render boundary (`effective_detail_panel_height` enforces the panel minimum and reserves shell/workspace minimum against window height), so persisted heights reclamp safely on smaller windows, both tabs share the height, and the global setting follows the serde-default round-trip pattern.
- Verified: workspace clippy clean, vibez-ui 270 passed / 0 failed.

Also noting for the record: with #78 merged, Phase 2 (Sections exist) is complete. Card 10 starts Phase 3.

alexanderwanyoike (commented):
## Re-review of 1b181be (reviewer: Claude, loop session)

Small responsive follow-up: when the detail panel and pad surface squeeze the Section area below 640px, the toolbar stacks its identity and controls vertically instead of truncating. Width flows down from the single `effective_perform_surface_width` computation rather than being re-derived, the breakpoint is a named constant with a threshold test, and vibez-ui passes 271 / 0 with clippy clean.

#79 remains clean; nothing else outstanding from my side on any open PR.

alexanderwanyoike:
HITL follow-up on commit 1b181be: narrowing Perform with Browser open exposed the Section toolbar controls consuming the full 460 px minimum pane width and collapsing the Section identity to one character per line. The toolbar now stacks below the identity under 640 px and returns inline when wider; no controls or tooltips were removed. Regression test passes, full vibez-ui suite passes (271/271), workspace clippy is clean, and the exact branch was visually verified at both narrow and wide widths. Evidence screenshots are in the dedicated target as pr79-perform-narrow-section.png and pr79-perform-wide-section.png.

PR #80, opened 2026-07-19
fix(ui): dispatch track shortcut once per press

alexanderwanyoike:
## Outcome

`Ctrl+T` and `Ctrl+Shift+T` create exactly one Track for one physical key press, including on X11 where auto-repeat arrives as synthetic release/press pairs.

## Scope

- one keyboard-adapter guard for the two Track-creation shortcuts
- ordinary shortcuts and intentional later Track shortcut presses remain unchanged
- no Track-domain or Arrange/Perform duplication

## Root cause

Iced 0.13 exposes no repeat flag. X11 may synthesize release/press pairs while a key remains held, so the original global shortcut handler dispatched every repeated keydown. The adapter now suppresses a press immediately following a synthetic release while allowing a genuine later press.

## Evidence

- red-first regression models an initial press, same-key keydown repeat, X11 synthetic release/press, and a later genuine press
- `cargo test --workspace -j4` (268 UI tests; 136 engine tests)
- exact branch PID/executable verified under Xvfb; held `Ctrl+Shift+T` produced one MIDI Track in `held-shortcut-fixed.png`/`instrumented-held.png`
- the first audio-enabled dogfood was independently interrupted by the known `SIGXCPU` device failure now isolated in the separate audio-disconnect PR

## Manual demo

1. Open an empty Arrange project.
2. Hold `Ctrl+Shift+T` long enough for keyboard repeat; exactly one MIDI Track should appear.
3. Release, then press it again; exactly one additional MIDI Track should appear.
4. Repeat with `Ctrl+T`; each physical press should add one Audio Track.

### DIFF
--- a/crates/vibez-ui/src/app/keyboard.rs
+++ b/crates/vibez-ui/src/app/keyboard.rs
@@ -12,6 +12,43 @@ use crate::domains::view::ViewMsg;
 
 use crate::message::Message;
 
+#[derive(Default)]
+pub(crate) struct EdgeShortcutState {
+    pressed_keys: std::collections::HashSet<String>,
+    released_at: std::collections::HashMap<String, std::time::Instant>,
+}
+
+impl EdgeShortcutState {
+    fn should_dispatch(
+        &mut self,
+        key_id: &str,
+        message: &Message,
+        occurred_at: std::time::Instant,
+    ) -> bool {
+        let edge_triggered = matches!(
+            message,
+            Message::Arrangement(ArrangementMsg::AddTrack | ArrangementMsg::AddInstrumentTrack)
+        );
+        if !edge_triggered {
+            return true;
+        }
+
+        let release_gap = self
+            .released_at
+            .get(key_id)
+            .map(|released_at| occurred_at.saturating_duration_since(*released_at));
+        let synthetic_repeat =
+            release_gap.is_some_and(|gap| gap <= std::time::Duration::from_millis(30));
+        let first_press = self.pressed_keys.insert(key_id.to_owned());
+        first_press && !synthetic_repeat
+    }
+
+    fn release(&mut self, key_id: &str, occurred_at: std::time::Instant) {
+        self.pressed_keys.remove(key_id);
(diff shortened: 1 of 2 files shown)


alexanderwanyoike (commented):
## PR 80 review (reviewer: Claude, loop session)

Reviewed under the same contract and strict bar. No findings; the duplication question this raises was the main thing I checked, and it passes.

- **Not a duplicate key tracker**: `EdgeShortcutState` superficially resembles `PerformState.active_computer_keys` (both track held keys by `runtime_key_id`), but their scopes are disjoint by construction: the Perform map only ever holds unmodified pad keys inside the Perform workspace, while the edge guard only ever engages for the two modifier-combo track-creation messages, which can never enter the pad path. Merging them would couple the Perform domain to router shortcut mechanics, which is worse than the two small structures. Placement in the keyboard adapter is right: X11 synthetic release/press pairs are an input-delivery quirk, not domain logic.
- **The 30ms heuristic is safe**: X11 synthetic pairs arrive with near-zero gaps, and no human can intentionally re-press within 30ms of a release, so the guard cannot swallow a genuine second press. The state machine handles the tricky sequences correctly (blocked synthetic press still registers as held; real release then re-press dispatches), and the test walks exactly that sequence.
- Edge-triggering is limited to an explicit message allowlist, so every other shortcut and all Perform pad gestures are untouched; `released_at` growth is bounded by distinct keys pressed.
- Verified: workspace clippy clean, vibez-ui 268 passed / 0 failed.

One tiny note, not a finding: if a third edge-triggered shortcut ever joins the allowlist, consider marking edge-triggered-ness on the message constructor side rather than growing the `matches!` in the guard, so the list cannot drift out of sync.

alexanderwanyoike:
Follow-up native reproduction invalidates this PR as the fix for the reported 1 → 2 → 4 → 8 behavior. The actual cause is duplicate shortcut ownership: every per-track TrackClipCanvas publishes a Track-creation message, so N existing Tracks create N more Tracks. The corrected fix and regression evidence are in #84. The key-repeat guard here did not cover the captured per-canvas path.

PR #81, opened 2026-07-19
fix(audio): survive output device disconnects

alexanderwanyoike:
## Outcome

A lost output device no longer leaves the Linux real-time callback in the ALSA hot loop that terminates Vibez with `SIGXCPU`. After CPAL reports a stream error, that stream stops processing DSP, demotes its callback thread, emits silence, and yields.

## Scope

- confined to `vibez-audio-io` stream health/recovery
- the normal real-time callback path is unchanged
- no automatic device reopening or UI device picker work

## Root cause evidence

Four Vibez coredumps are `SIGXCPU` (including the exact Card 08 dogfood binary). The audio callback had been promoted through `audio_thread_priority`, whose Linux implementation sets `RLIMIT_RTTIME` to roughly one callback budget (about 11.6 ms for 512 frames at 44.1 kHz). ALSA can continue callbacks without blocking after a USB/device loss, exhausting that budget and causing the kernel signal.

## Evidence

- red-first regression proves a stream error permanently latches subsequent callbacks into silence/yield mode
- `cargo test --workspace -j4` (31 audio-I/O tests; 267 UI tests; 136 engine tests)
- exact branch executable and PID verified under Xvfb; normal stream promoted and stayed alive through smoke dogfood
- a separate exact shortcut-branch dogfood reproduced the pre-fix `SIGXCPU`, strengthening the crash diagnosis

## Manual demo

This is optional hardware sign-off: play a project through a removable USB/audio output, disconnect that output, and confirm Vibez remains open rather than disappearing. Audio from that failed stream should become silent. Device auto-reopen is intentionally out of scope.

### DIFF
--- a/crates/vibez-audio-io/src/audio_stream.rs
+++ b/crates/vibez-audio-io/src/audio_stream.rs
@@ -5,7 +5,9 @@
 //! inside the real-time audio callback.
 
 use std::fmt;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 use cpal::{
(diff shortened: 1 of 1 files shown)


alexanderwanyoike (commented):
## PR 81 review (reviewer: Claude, loop session)

Reviewed under the same contract. No findings; this is a well-confined fix for a real production crash (the SIGXCPU coredumps include the Card 08 dogfood binary, so this was actively biting).

- **The latch is correct lock-free practice**: an `AtomicBool` set from cpal's error callback with Release, read in the audio callback with Acquire; one extra atomic load is the only cost on the healthy path, which keeps the "normal real-time callback unchanged" claim honest.
- **Sleeping in an audio callback is normally forbidden, and the exception is justified here**: once ALSA reports the device gone, the stream is unrecoverable and ALSA can hot-loop the callback, which is exactly what burned the RLIMIT_RTTIME budget and produced SIGXCPU. Demoting the thread from real-time first (once, via the `take()`), then silence plus a 10ms yield, is the right sequence, and the inline comment documents why the rule is suspended. The latch test covers the state machine.
- Scope is genuinely confined to `vibez-audio-io`; no engine or UI changes smuggled in.

One observation for a future card, not a finding: after the latch trips, Vibez plays silence with no UI indication that the device died; the producer sees a working app that makes no sound. A stream-health event surfaced to the status bar (and eventually the Settings device picker) would complete the story, and pairs naturally with the deferred automatic-reopen work the PR description already excludes.

PR #82, opened 2026-07-19
feat(perform): play one Section

alexanderwanyoike:
## Summary

- adapt authored Sections into the Card 09 resident playback-source seam for every Project Track, including intentional empty sources
- launch Sections immediately from computer pads or a distinct pointer affordance while keeping editing selection independent
- keep Section playhead/playing-pad truth engine-owned, preserve shared channel devices and tails, release sounding notes on switches, and restore Arrange playback after stop or a non-looping end
- keep the callback switch path allocation-, lock-, I/O-, load-, and destruction-free by swapping prepared boxed sources and returning retired ownership to the UI thread

## Verification

- `cargo clippy -j 4 --workspace -- -D warnings`
- `cargo test -j 4 --workspace`
- focused TDD regressions cover immediate switching, shortened loop boundaries, Arrange-loop isolation, non-looping return to Arrange, silent absent tracks with continuing effect tails, and safe note release
- all changed Rust files remain below 1,000 lines

## Native dogfood

Exact commit `a73e6f6` was built and run from its dedicated target/worktree on isolated Xvfb display `:98`. `/proc` and the X11 window PID confirmed the branch executable. Pointer launch started Section 01; keyboard `2` then started Section 02 while Section 01 remained selected, confirming engine-driven playing state and independent editing selection. Evidence is in the dedicated Cargo target as `card10-xvfb-*.png`.

Audible sign-off with two contrasting authored Sections remains a human merge gate. Concise steps are included in the card delivery evidence and will be posted on this PR.

Card 10 only. Quantization/residency (Card 11), later Perform work, hardware cards, and the separate Arrange add-track/styling bug are out of scope.

### DIFF
--- a/crates/vibez-engine/src/commands.rs
+++ b/crates/vibez-engine/src/commands.rs
@@ -7,6 +7,8 @@ use vibez_core::track::DrumPadState;
 use vibez_dsp::effect::AudioEffect;
 use vibez_instruments::Instrument;
 
+use crate::playback_source::PreparedSectionPlaybackSource;
+
 /// Quantization policy for the next Browser Audition start.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AuditionSync {
@@ -30,6 +32,8 @@ pub enum EngineCommand {
     Seek(u64),
     /// Change the project tempo.
     SetBpm(f64),
+    /// Immediately activate a complete resident Section playback source.
+    LaunchSection(Box<PreparedSectionPlaybackSource>),
     /// Load decoded audio into the engine for playback (legacy single-file).
     LoadAudio(Arc<DecodedAudio>),
     /// Remove any loaded audio from the engine (legacy single-file).
--- a/crates/vibez-engine/src/commands.rs
+++ b/crates/vibez-engine/src/commands.rs
@@ -34,6 +34,10 @@ pub enum EngineCommand {
     SetBpm(f64),
     /// Immediately activate a complete resident Section playback source.
     LaunchSection(Box<PreparedSectionPlaybackSource>),
+    /// Replace the currently active Section's resident source in place.
+    /// The engine preserves its local playhead and returns ownership of the
+    /// displaced source through `SectionSourceRefreshed`.
+    RefreshSection(Box<PreparedSectionPlaybackSource>),
     /// Load decoded audio into the engine for playback (legacy single-file).
     LoadAudio(Arc<DecodedAudio>),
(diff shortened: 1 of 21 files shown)


alexanderwanyoike (commented):
## Card 10 review (reviewer: Claude, loop session)

Reviewed against `perform/cards/10-play-one-section.md`, the delivery contract, and the Card 09 requirements I recorded on PR #77. This is the riskiest slice so far and it landed clean: no fix requested.

### The two carried requirements are properly met
- **Off-thread drop, elegantly**: `LaunchSection` swaps each track's boxed source in place, so the SAME `PreparedSectionPlaybackSource` owner that carried the new content out carries the displaced content back through `SectionTransitioned { retired }`, where the UI thread explicitly drops it. The full-event-queue edge `std::mem::forget`s rather than dropping in the callback, trading a rare one-source leak for real-time safety, with the reasoning documented inline. No allocation or destruction on the audio thread anywhere in the path.
- **Seam validated by the second adapter**: Sections flow through the Card 09 `PreparedPlaybackSource` boundary per track; no duplicate rendering machinery appeared.

### Independently verified
- `cargo clippy --workspace -- -D warnings`: clean; `cargo test --workspace`: 690 passed, 0 failed.
- **Intentional silence at the right layer**: `prepare_playback_source` emits a source for every Project Track, empty for absent ones, so silence is expressed without resetting shared channels and tails decay; the swap never touches instruments/effects.
- **Engine truth**: pad playing state is set only from `SectionTransitioned` and cleared on `PlaybackStopped`; `SectionPlaybackPosition` carries the engine-owned local playhead. No UI-timer inference.
- **The five engine tests hit every criterion**: immediate switch with displaced-source return, mid-callback wrap at a shortened non-destructive boundary, Arrange-loop independence from Section time (the three-loop-concepts rule), non-looping Section ending back to the Arrange source, and note release on switch.
- **File discipline**: `engine.rs` was proactively split (836 lines, with audition extracted to its own 249-line module) instead of brushing the 1000 limit.
- `docs/ARCHITECTURE.md` updated; scope bounded (no quantization, queueing, residency preloading, banks, or BPM lock, all Card 11).

### For Alex's gate: this is THE feel card
Phase 3's whole promise is that launching feels tight. Before merging, play with it: launch from a key, launch from the pointer affordance, switch between two contrasting Sections mid-bar, and listen for the tail carrying over a transition while the absent track goes silent. If the immediate switch feels mushy or glitchy here, everything Card 11 builds on top inherits it.

PR #85, opened 2026-07-19
Fix VST3 instruments as MIDI clip sources

alexanderwanyoike:
### Outcome
- reads VST3 PClassInfo2 subcategories so Instrument components enter the Track instrument slot and Fx components remain inserts
- exposes queued note-on/note-off events to IAudioProcessor::process through a callback-lifetime IEventList
- preserves within-block frame offsets and orders note-off before same-offset retriggers
- automatically refreshes legacy plugin catalogs once, so existing installs receive the corrected VST3 classifications without a hidden manual rescan
- leaves CLAP, built-in instruments, VST3 effects, plugin state, and Perform Card 11 untouched

### Root cause and follow-up
The first native sign-off failed because the persisted pre-fix catalog still classified Dexed, Surge XT, and Vital VST3 as Effects. The live trace proved the UI therefore inserted Dexed and Surge as effect devices; MIDI never reached an instrument. Fresh scanning already classified all three correctly.

A versioned plugin catalog now requests one sandboxed startup refresh when scanner semantics change, persists the current revision after completion, and does not rescan on subsequent launches.

### TDD and verification
- red: scanner had no category contract and every VST3 was hard-coded Effect
- red: VST3 process inputEvents was null and timed note offsets were not stored
- red: a legacy catalog had no way to request and remember exactly one migration refresh
- green: complete workspace suite, strict workspace Clippy, format, and diff checks
- isolated legacy-catalog startup migrated Dexed, Surge XT, and Vital from Effect to Instrument, populated their vendors, and left the catalog byte-identical on a second launch
- direct installed-plugin probes produced non-zero audio from timed MIDI: Dexed peak 0.13078251; Surge XT peak 0.33702826

### Manual sign-off
1. Start this exact branch with the existing pre-fix catalog; wait for the automatic scan to finish.
2. Remove any Dexed or Surge devices previously added as effects.
3. Add Dexed VST3 and Surge XT VST3 again; verify each occupies the instrument card.
4. Play an authored MIDI clip and confirm each selected VST3 is audible.
5. Add a VST3 effect separately and confirm it remains an insert.

The large VST3 instance module remains intentionally local per the canonical refactoring guide; this PR does not split its unsafe FFI contract.

### DIFF
--- a/crates/vibez-plugin-host/examples/scan_dir.rs
+++ b/crates/vibez-plugin-host/examples/scan_dir.rs
@@ -21,7 +21,7 @@ fn main() {
     let settings = PluginSettings {
         extra_scan_paths: vec![dir],
         scan_default_paths: false,
-        cache: Vec::new(),
+        ..PluginSettings::default()
     };
 
     if sandboxed {
@@ -49,7 +49,7 @@ fn main() {
         let single = PluginSettings {
             extra_scan_paths: vec![path.clone()],
             scan_default_paths: false,
-            cache: Vec::new(),
+            ..PluginSettings::default()
         };
         // scan_plugins only walks directories, so scan the parent dir
         // trick doesn't isolate; call the per-format scanners directly.
--- a/crates/vibez-plugin-host/src/settings.rs
+++ b/crates/vibez-plugin-host/src/settings.rs
@@ -4,12 +4,16 @@ use serde::{Deserialize, Serialize};
 
 use crate::info::PluginInfo;
 
+const CURRENT_CACHE_REVISION: u32 = 1;
+
 /// Plugin host settings, persisted to `~/.config/vibez/plugins.json`.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PluginSettings {
     pub extra_scan_paths: Vec<PathBuf>,
     pub scan_default_paths: bool,
     pub cache: Vec<PluginInfo>,
+    #[serde(default)]
+    pub cache_revision: u32,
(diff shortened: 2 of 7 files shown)


alexanderwanyoike (commented):
## Card 26 review (reviewer: Claude, loop session)

Reviewed against the new `perform/cards/26-route-midi-clips-to-vst3-instruments.md`. No fix requested.

- **Subcategory routing**: the scanner now reads `PClassInfo2` subcategories so Instrument components land in the Track instrument slot while Fx components stay inserts, closing the gap where VST3 instruments could not source MIDI clips.
- **Event delivery is correct MIDI practice**: queued note-ons/offs reach `IAudioProcessor::process` through a callback-lifetime `IEventList`, within-block frame offsets preserved, and note-off ordered before a same-offset retrigger so repeated pitches do not hang.
- **`instance.rs` at 1339 lines is within its documented exemption**: the refactoring guide explicitly excludes `vst3_host/instance.rs` from the 1000-line rule because its unsafe FFI contract is intentionally local. Growth of 197 lines for the event list plumbing stays inside that rationale; flagging only so the exemption stays a conscious choice.
- Verified: workspace clippy clean; vibez-plugin-host + vibez-ui suites 323 passed / 0 failed.

Worth a hardware-adjacent dogfood note for Alex: this is the card that makes finger-drumming into Dexed/Surge possible when Card 12 lands, so a quick native check that a real VST3 instrument sounds from a MIDI clip would validate the whole chain early.

PR #86, opened 2026-07-19
Perform: queue Sections on quantized boundaries

alexanderwanyoike:
### Outcome
- prepares Section residency through the shared cancellable TrackedRequest mechanism from Card 25
- queues Immediate, 1 Beat, 1 Bar, and End of Section policies in the engine
- swaps resident sources exactly at engine-owned boundaries, including mid-callback, and reports sample timestamps
- drives PLAYING, QUEUED, and pending-boundary UI state only from engine events
- adds independent bracket-key banks for Sections, Track Mutes, and the retained Instrument shell
- locks project BPM throughout Perform playback and preserves live Track mute state across transitions

### Ownership and RT safety
- UI/background code prepares decoded clip owners before QueueSection is sent
- requeue aborts stale UI work and returns displaced resident owners for UI-thread destruction
- the callback swaps pointers only; no loading, locks, I/O, or resident-owner drops
- the Card 10 immediate LaunchSection and playing-Section refresh seams remain intact

### TDD and verification
- red/green engine coverage for all four policies, mid-buffer output and timestamp, stale requeue retirement/latest-wins, BPM lock, and mute persistence
- red/green UI coverage for independent banks/no engine commands, engine-event playback truth, BPM rejection, and shared request cancellation/stale-result behavior
- 148 vibez-engine tests pass
- 283 vibez-ui tests pass
- cargo fmt --all -- --check
- cargo check --workspace
- cargo clippy --workspace --all-targets with -D warnings passes when allowing only three pre-existing origin/dev Rust 1.96 lints
- cargo test --workspace
- git diff --check

Strict all-target Clippy without those allowances remains red on unchanged origin/dev files: host_context.rs, engine_section_playback_tests.rs, and ui_settings.rs. None are touched here. All touched Rust files remain under 1000 lines.

### Exact-branch native dogfood
- executable, cwd, PID, and X11 window all verified against this worktree and target on Xvfb only
- created two Sections; first changed to PLAYING only after its engine event
- queued the second at 1 Bar; pad changed to QUEUED and header displayed the engine-reported absolute beat boundary
- the isolated Xvfb audio backend runs callbacks faster than wall clock and floods position events, so it is not accepted as timing/audibility evidence; deterministic engine tests cover exact switching and the real native audio pass remains the human gate

### Manual sign-off
1. Prepare two contrasting looping Sections and set the second to 1 Bar.
2. Launch the first, then launch the second mid-bar. Confirm the second says QUEUED with a pending beat and becomes PLAYING exactly on the bar.
3. Repeat with Immediate, 1 Beat, and End of Section.
4. Mute a Track before a transition and confirm it stays muted afterward until reversed.
5. While Perform plays, try typing and nudging BPM; it must remain unchanged. Stop, then confirm BPM edits work.
6. Create a Section in slot 17, use ] to reach bank 2 and [ to return, switch modes, and confirm each mode remembers its own bank without interrupting playback.

Out of scope and intentionally absent: Instrument performance behavior, recording, Capture, Card 12+, and the separate Arrange add-track styling bug.

### DIFF
--- a/crates/vibez-core/src/lib.rs
+++ b/crates/vibez-core/src/lib.rs
@@ -6,5 +6,6 @@ pub mod effect;
 pub mod id;
 pub mod midi;
 pub mod onset;
+pub mod perform;
 pub mod time;
 pub mod track;
--- a/crates/vibez-core/src/perform.rs
+++ b/crates/vibez-core/src/perform.rs
@@ -1,0 +1,36 @@
+use serde::{Deserialize, Serialize};
+
+/// Musical boundary at which a resident Section launch becomes effective.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SectionLaunchQuantization {
+    Immediate,
+    OneBeat,
+    #[default]
+    OneBar,
+    EndOfSection,
+}
+
+impl SectionLaunchQuantization {
+    pub const ALL: [Self; 4] = [
+        Self::Immediate,
+        Self::OneBeat,
+        Self::OneBar,
+        Self::EndOfSection,
+    ];
+
+    pub const fn label(self) -> &'static str {
+        match self {
+            Self::Immediate => "Immediate",
+            Self::OneBeat => "1 Beat",
+            Self::OneBar => "1 Bar",
+            Self::EndOfSection => "End of Section",
+        }
+    }
+}
+
+impl std::fmt::Display for SectionLaunchQuantization {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str(self.label())
+    }
+}
--- a/crates/vibez-engine/src/commands.rs
+++ b/crates/vibez-engine/src/commands.rs
@@ -3,6 +3,7 @@ use vibez_core::audio_buffer::DecodedAudio;
 use vibez_core::effect::EffectType;
 use vibez_core::id::{ClipId, EffectId, TrackId};
 use vibez_core::midi::{InstrumentKind, MidiNote};
+use vibez_core::perform::SectionLaunchQuantization;
 use vibez_core::track::DrumPadState;
 use vibez_dsp::effect::AudioEffect;
 use vibez_instruments::Instrument;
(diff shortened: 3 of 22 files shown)


alexanderwanyoike (commented):
## Card 11 review (reviewer: Claude, loop session)

Reviewed against `perform/cards/11-queue-sections-with-quantized-transitions.md`, the delivery contract, and the requirements carried from the Card 10 and Card 25 reviews. One fix requested; everything substantive is right.

### The carried requirements are all properly met
- **TrackedRequest for residency, as required**: `begin_section_residency` runs preparation off-thread via `spawn_blocking`, tracked by the Card 25 `TrackedRequest` with `begin()`/`attach()`; the completion handler rejects stale tokens via `finish(request_id)`, re-launches replace in-flight preparation, transport Stop cancels it, and a content edit to a queued Section re-runs residency. No fifth mechanism appeared.
- **Off-thread drop extended to the queue**: `SectionQueued { retired }` and `SectionQueueCancelled { retired }` return displaced prepared sources to the UI thread, with the same forget-on-full-queue discipline; `requeue_returns_stale_residency_and_only_latest_section_transitions` covers it.
- **Mid-buffer quantized boundaries with engine timestamps**: `process_section_multitrack` split-renders the old Section up to the boundary sample, activates exactly there, renders the new Section for the remainder; `one_beat_transition_lands_mid_buffer_with_engine_timestamp` asserts the timestamp, and OneBar/EndOfSection have exact-boundary tests.
- **BPM lock and mute persistence**: tempo changes are rejected while a Section plays or is queued (with a status message via `TempoRejected`), and `perform_bpm_is_locked_and_live_mute_survives_transition` covers both criteria in one engine test.

### Also verified
- `cargo clippy --workspace -- -D warnings`: clean; `cargo test --workspace`: 706 passed, 0 failed.
- `SectionLaunchQuantization` moved to `vibez-core::perform` (needed by the engine) with vibez-project re-exporting rather than duplicating.
- Queued pad state and pending boundary display come from `SectionQueued`/`SectionTransitioned` events, cleared on `PlaybackStopped`; no UI-timer inference.
- `docs/ARCHITECTURE.md` updated.

### Finding
1. **(fix requested, delivery contract)** `views_perform.rs` is at **1021 lines**, over the hard 1000-line limit. The contract says split before crossing, and this file has been flagged twice before (965 at Card 06, split at Card 07 brought it down). The natural cut: move the Sections-mode pad rendering (pad faces, queued/playing badges, launch affordance) into a `views_perform_pads.rs` sibling, mirroring the existing `views_perform_sections.rs` split.

Will re-review the delta when the split lands. Alex: this card completes the launch loop, so the feel pass now covers queue-then-switch timing, not just immediate launch.

alexanderwanyoike:
Re-reviewed at 19460cd. The file-size fix request is resolved: views_perform.rs is down to 614 lines with the Pad Surface rendering extracted into views_perform_pads.rs (423 lines), matching the existing sibling-module pattern (views_perform_playhead, views_perform_sections). I verified the extraction is a pure move: every removed line reappears in the new module, with only the import lists re-split between the two files, and shared helpers exposed as pub(super).

Independently verified at this head: cargo fmt clean, workspace clippy with -D warnings clean, all 284 UI tests pass, and all eight CI jobs are green. No open findings remain from my side; merge stays behind Alex's sign-off (visual check of the new bank buttons was still pending).

PR #87, opened 2026-07-19
Fix looped MIDI loudness growth after bar 2

alexanderwanyoike:
## Summary
- emit each looped MIDI note-on and note-off exactly once when its beat boundary is crossed
- apply the same boundary rule to built-in per-frame and plugin batch rendering
- add production-rate four-bar regressions for both rendering paths

## Root cause
At 44.1 kHz, the floating-point one-sample event window became true on two adjacent frames after the second one-bar wrap. Identical kicks were therefore triggered twice from bar 3 onward.

## Verification
- red: both paths reported note-ons per bar as [8, 8, 16, 16]
- green: both paths report [8, 8, 8, 8]
- exact HM BD 02.wav render: peak 0.41210467 and RMS 0.17934328 in every bar
- cargo test --workspace
- cargo clippy --workspace -- -D warnings
- cargo fmt --all -- --check
- git diff --check

## Scope
Isolated Card 27 fix only. No Card 11, VST routing, Explorer, or UI changes. Not merged pending explicit approval and native playback sign-off.

### DIFF
--- a/crates/vibez-engine/src/engine_stuck_note_tests.rs
+++ b/crates/vibez-engine/src/engine_stuck_note_tests.rs
@@ -12,6 +12,7 @@ type NoteEventLog = Arc<Mutex<Vec<(bool, u8)>>>;
 /// Instrument that records every note event it receives.
 struct SpyInstrument {
     events: NoteEventLog,
+    batch_render: bool,
 }
 
 impl vibez_instruments::Instrument for SpyInstrument {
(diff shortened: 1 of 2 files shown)


alexanderwanyoike (commented):
## PR 87 review (reviewer: Claude, loop session)

No findings. This is the right fix shape: the fragile absolute-window test (`diff >= 0 && diff < beat_step`, which double-fires when float accumulation makes the window true on two adjacent frames after a loop wrap) is replaced by an edge-crossing predicate against the previous frame's position, with half-open semantics (`previous < event <= current`) guaranteeing exactly-once emission. One shared `crossed_beat` helper serves both the per-frame and plugin-batch paths instead of two fixed copies, which is exactly the reuse bar. The wrap-frame arm retains its retrigger-sounding-notes semantics. Production-rate four-bar regressions cover both paths (red before, green after per the PR body); vibez-engine suite 145 passed / 0 failed, workspace clippy clean.

PR #88, opened 2026-07-20
fix(browser): select visible Remote folders directly

alexanderwanyoike:
## Summary

- activate the Remote Place inside the existing `SelectRemoteFolder` domain transition
- cover one-click selection of both a visible Dropbox connection root and nested folder from Local results
- keep disclosure controls navigation-free and preserve query, scope, selection reset, catalog, and bounded-result behavior

This PR implements Browser Card 15 only. The corresponding Local-return defect remains isolated in Card 16.

## Diagnosis and TDD

The deterministic regression reproduced on fresh `origin/dev`: after `SelectRemoteFolder("/megalodon/drums")`, the exact path, expansion, selected-folder scope, selection reset, and results bound were correct, but `SampleBrowserMode` remained `Local`. The test was observed RED (`left: Local`, `right: Remote`) before the one-line fix and GREEN afterward.

## Verification

- `cargo check --workspace -j 4`
- `cargo clippy --workspace -- -D warnings`
- `git diff --check`
- exact-head Xvfb dogfood at `d9a111e`: from Local, one click on visible `Art` activated and highlighted that Remote folder with its children in Results; from Local, one click on `Alex’s Dropbox` activated and highlighted the connection root; the Art disclosure control collapsed without navigating

All Cargo commands used the lane target and shared global lock.

### DIFF
--- a/crates/vibez-ui/src/domains/browser.rs
+++ b/crates/vibez-ui/src/domains/browser.rs
@@ -382,6 +382,7 @@ impl BrowserState {
                 self.remote.connection_expanded = !self.remote.connection_expanded;
             }
             BrowserMsg::SelectRemoteFolder(path) => {
+                self.mode = SampleBrowserMode::Remote;
                 self.remote.current_path = path;
                 if !self.remote.current_path.is_empty() {
                     self.remote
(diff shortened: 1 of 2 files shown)


alexanderwanyoike:
Reviewed at d9a111e against Browser Card 15. Clean: the fix is the minimal one-line mode activation inside the existing SelectRemoteFolder transition, exactly where the card's diagnosis pointed, with regression coverage for connection-root and nested-folder selection from Local mode plus preserved query/scope/selection/bounded-results behavior. Independently verified at this head: fmt clean, all 286 UI tests pass, all eight CI jobs green. No findings; merge stays behind Alex's approval.

PR #89, opened 2026-07-20
fix(browser): select visible Local folders directly

alexanderwanyoike:
## Summary
- activate the Local Place whenever a visible All Roots, Local Root, or nested folder node is selected
- keep disclosure controls expansion-only and leave watcher reconciliation on the path-only helper
- add regression coverage for exact-node activation and preserved Browser context

## Verification
- `git diff --check`
- workspace Clippy clean with only four independently reproduced Rust 1.96 baseline lints allowed; strict Clippy remains red only in untouched `origin/dev` files
- exact-head isolated Xvfb dogfood for disclosure, All Roots, Local Root, and nested folder navigation

Private Browser Card 16 only. Targets `dev`; do not merge without explicit approval.

### DIFF
--- a/crates/vibez-ui/src/domains/browser.rs
+++ b/crates/vibez-ui/src/domains/browser.rs
@@ -126,8 +126,14 @@ impl BrowserState {
                 self.select_local_folder(folder);
             }
             BrowserMsg::ToggleLocalFolder(folder) => {
-                if !self.expanded_local_folders.remove(&folder) {
-                    self.expanded_local_folders.insert(folder);
+                let should_expand = !self.expanded_local_folders.remove(&folder);
+                if should_expand {
+                    self.expanded_local_folders.insert(folder.clone());
+                }
+                self.mode = SampleBrowserMode::Local;
+                self.select_local_folder(Some(folder.clone()));
+                if !should_expand {
+                    self.expanded_local_folders.remove(&folder);
                 }
             }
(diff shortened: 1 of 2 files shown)


alexanderwanyoike:
Reviewed at 6ec18f1 against Browser Card 16 (including the revised disclosure acceptance criterion from Alex's live review). Correct: SelectLocalFolder now activates the Local Place atomically, and the disclosure toggle keeps its expand/collapse action while also activating the folder, matching the updated card. Alex's combined 15/16 human retest at 27ee2bd is on record. Independently verified at this head: fmt clean, workspace clippy with -D warnings clean, all 287 UI tests pass, all eight CI jobs green. No fix requests. Two notes:

1. Observation, not blocking: the ToggleLocalFolder insert-select-remove sequence exists only because BrowserState::select_local_folder couples selection with force-expansion. If a third caller ever needs selection without expansion, split the expansion side effect out rather than adding another compensating removal.
2. For Alex: Cards 15 and 16 now encode opposite disclosure semantics. Remote chevrons only expand/collapse and do not navigate (Card 15 criterion), while Local chevrons expand/collapse AND activate the folder (Card 16's revised criterion). The Remote tree therefore still has the row-vs-chevron trap that the Local human test exposed: clicking a Remote chevron while Local results are active leaves Remote inactive. If the retest felt fine this may be intentional, but if not, aligning Remote disclosure with the new Local behavior would be a small follow-up card.

PR #90, opened 2026-07-20
feat(perform): play instrument targets from pads

alexanderwanyoike:
## Summary

- play Perform Instrument pads directly through the engine command path with paired note-on/off state
- share the project-wide selected playable MIDI track between the Shift target overlay, on-screen picker, Section editor, and unmatched MIDI input
- add independent Instrument banking, bottom-left pitch/target order, and globally persisted fixed computer-key velocity
- keep held notes paired across target changes, mode changes, and queued Section transitions

## Verification

- full workspace test suite passed (UI 292 tests; engine 151 tests; remaining crates and doc tests passed)
- workspace Clippy passed with `-D warnings`
- `git diff --check origin/dev...HEAD`
- exact-head isolated Xvfb/null-sink dogfood recorded under `/home/alexander/Code/Apps/vibez-targets/card-12/card12-dogfood/evidence`
  - built-in Subtractive Synth, Dexed VST3, and Surge XT VST3 all produced non-silent pad audio
  - Shift retargeting and the on-screen target picker both exercised
  - velocity changed from 100 to 64 immediately and remained 64 after a full app restart
  - Section 02 launched at sample 85,355 while Bass C2 remained held across the queued one-bar transition, then released to digital silence without a stuck note

## Scope

- Card 12 only; Cards 13, 14, and 28 remain out of scope
- no merge requested

### DIFF
--- a/crates/vibez-engine/src/engine_section_playback_tests.rs
+++ b/crates/vibez-engine/src/engine_section_playback_tests.rs
@@ -419,6 +419,62 @@ fn switching_sections_releases_notes_from_the_previous_source() {
 }
 
 #[test]
+fn live_instrument_note_releases_after_a_section_transition() {
+    let (mut engine, mut commands, _events) = AudioEngine::new();
+    let track_id = TrackId::new();
+    let note_events = Arc::new(Mutex::new(Vec::new()));
+    commands
+        .push(EngineCommand::AddMidiTrack(track_id, "MIDI".into()))
+        .unwrap();
+    commands
+        .push(EngineCommand::SetPluginInstrument {
+            track_id,
+            instrument: Box::new(NoteLogInstrument(Arc::clone(&note_events))),
+        })
+        .unwrap();
+    commands
+        .push(EngineCommand::LaunchSection(Box::new(
+            PreparedSectionPlaybackSource::new(
+                SectionId::new(),
+                4.0,
+                true,
+                vec![(track_id, PreparedPlaybackSource::default())],
+            ),
+        )))
+        .unwrap();
+    commands
+        .push(EngineCommand::ExternalNoteOn {
+            track_id,
+            pitch: 48,
+            velocity: 100,
+        })
+        .unwrap();
+    let mut output = [0.0; 1];
+    engine.process(&mut output, 1);
+
+    commands
+        .push(EngineCommand::LaunchSection(Box::new(
+            PreparedSectionPlaybackSource::new(
+                SectionId::new(),
+                4.0,
+                true,
+                vec![(track_id, PreparedPlaybackSource::default())],
+            ),
+        )))
+        .unwrap();
+    engine.process(&mut output, 1);
+    commands
+        .push(EngineCommand::ExternalNoteOff {
+            track_id,
+            pitch: 48,
+        })
+        .unwrap();
+    engine.process(&mut output, 1);
+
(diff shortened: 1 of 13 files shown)


alexanderwanyoike:
Reviewed at 42608cb against Card 12 and the delivery contract.

**Verified working:**
- Live notes reuse the existing ExternalNoteOn/ExternalNoteOff engine path shared with hardware MIDI input; no parallel note primitive was introduced, and no engine source changes were needed. Gestures reach the engine from the keyboard event handler, off the 60 fps tick.
- Each held key stores its original target track and pitch (ActiveInstrumentNote), so octave shifts, retargeting, mode changes, and Section transitions cannot redirect a note-off. Covered by held_note_release_stays_paired_when_range_target_and_mode_change and the new engine test live_instrument_note_releases_after_a_section_transition.
- The fixed-velocity global setting is done right: serde default 100 for existing configs, round-trip test, applied on startup in app init, immediate-effect slider in Perform settings, persisted via the existing ui.json path.
- is_playable_midi_target() unifies the playable-track predicate across the domain, actions, and views, replacing the previous inline kind.is_midi() && has_instrument checks (including one pre-existing site in actions.rs). Target selection reuses project-wide track selection rather than a second selection value.
- Shift overlay semantics: ModifiersChanged forwarding always lets shift releases through even when a widget captured the event, so the overlay cannot stick after focus moves into a text field; showing is gated on workspace visibility.
- Octave clamping to MIDI 0-127 with the -3..+6 range; ARCHITECTURE.md updated; all touched files under 1,000 lines (heads-up: domains/perform.rs is at 972 and perform_tests.rs at 996, so the next Perform card will likely need to split one or both).
- Independently ran fmt, workspace clippy -D warnings, and the full workspace suite (722 tests) at this head: all clean. All eight CI jobs green.

**One fix requested — stuck note on window focus loss:**
Computer-key note-offs depend on a KeyReleased event. If the window loses OS focus while a pad key is held (Alt-Tab, a popup stealing focus, workspace switch), the release goes to the other window and the note sounds until the player re-presses and releases the same key. Hardware MIDI is immune to this; the computer-key path introduced here is not, and in a live-performance feature a stuck note is exactly the failure this card's safe-release criteria exist to prevent. The fix is small and uses an existing mechanism: the subscription in app/mod.rs already matches iced::Event::Window for Resized; add iced::window::Event::Unfocused, and on it drain active_computer_keys, sending ExternalNoteOff for every entry carrying an ActiveInstrumentNote (also a sensible place to clear the Shift overlay, since a ModifiersChanged clearing it is not guaranteed while unfocused). A domain test that simulates focus loss with two held notes and asserts both matching note-offs would lock the behavior in.

alexanderwanyoike:
Addressed the focus-loss blocker in `8e1f1dd`.

- The iced window subscription now routes `window::Event::Unfocused` into Perform.
- Perform drains every held computer key and sends `ExternalNoteOff` with the original track and pitch for each note.
- The Shift target overlay is cleared at the same boundary.
- A two-held-note domain regression verifies both matching note-offs, cleared pad state, and overlay cleanup.
- The new regression is split into its own test module; `perform.rs` remains at 988 lines and `perform_tests.rs` at 996.

Local verification: formatting and diff checks pass; the full workspace suite passes (294 `vibez-ui` tests); workspace Clippy passes with only the three pre-existing Rust 1.96 baseline lints allowed. Cross-platform CI is now running on the exact head.

PR #91, opened 2026-07-20
Perform Card 13: add Full Level and 16 Levels

alexanderwanyoike:
## Summary

- add Full Level as a reversible effective-velocity transform shared by Instrument input adapters
- add data-driven Pitch and Velocity 16 Levels assignments with editable ranges
- add remembered/chooseable source lifecycle that clears when the Instrument Target changes
- add integrated Perform controls and per-pad pitch/velocity feedback
- document the live note-transform architecture

## Verification

- `cargo test --workspace -j 4` (307 vibez-ui tests, 151 engine tests, complete workspace and doc tests)
- strict workspace Clippy passes for this change with three known Rust 1.96 baseline lints in untouched files explicitly allowed (`unnecessary_cast`, `replace_box`, `items_after_test_module`)
- isolated Xvfb + Pulse null-sink dogfood: Full Level, Pitch and Velocity assignments, live range edits, Source reassignment, and target-change clearing
- velocity pads 1 and 16 produced distinct recorded audio peaks (-29.30 dB vs -4.87 dB)

## Scope

Card 13 only. Note Repeat, Swing, and recording remain out of scope.

Do not merge without Alex’s explicit approval.

### DIFF
--- a/crates/vibez-ui/src/app/actions.rs
+++ b/crates/vibez-ui/src/app/actions.rs
@@ -120,7 +120,10 @@ impl App {
         }
         if let Some(track_id) = action.select_project_track {
             self.state.arrangement.selected_track = Some(track_id);
-            self.state.perform.sync_instrument_target(Some(track_id));
+            self.state.perform.sync_instrument_target_from_selection(
+                Some(track_id),
+                &self.state.project_tracks.tracks,
+            );
             if self.state.perform.selected_section.is_some() {
                 self.state
                     .perform
--- a/crates/vibez-ui/src/app/actions.rs
+++ b/crates/vibez-ui/src/app/actions.rs
@@ -120,6 +120,7 @@ impl App {
         }
         if let Some(track_id) = action.select_project_track {
             self.state.arrangement.selected_track = Some(track_id);
+            self.state.perform.sync_instrument_target(Some(track_id));
             if self.state.perform.selected_section.is_some() {
                 self.state
                     .perform
--- a/crates/vibez-ui/src/app/mod.rs
+++ b/crates/vibez-ui/src/app/mod.rs
@@ -160,6 +160,7 @@ mod views_devices;
 mod views_mixer;
 mod views_overlays;
 mod views_perform;
+mod views_perform_instrument;
 mod views_perform_pads;
 mod views_perform_playhead;
 mod views_perform_sections;
--- a/crates/vibez-ui/src/app/update_timeline.rs
+++ b/crates/vibez-ui/src/app/update_timeline.rs
@@ -62,7 +62,10 @@ impl App {
                     .section_editor
                     .editor_mut()
                     .selected_track = Some(*track_id);
-                self.state.perform.sync_instrument_target(Some(*track_id));
+                self.state.perform.sync_instrument_target_from_selection(
+                    Some(*track_id),
+                    &self.state.project_tracks.tracks,
+                );
                 return ArrangementAction::default();
             }
         }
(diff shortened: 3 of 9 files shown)


alexanderwanyoike:
Reviewed at 644374c against Card 13 and the delivery contract.

**Verified working:**
- The extensibility criterion is genuinely met: SIXTEEN_LEVELS_PARAMETERS is a data-driven descriptor table (label, bounds, default range, owns_velocity, apply fn), the runtime path maps through descriptor.apply with no pitch/velocity branch pair, and descriptors_distribute_values_without_parameter_specific_pad_logic locks it in. Adding a filter/decay parameter later is one table row plus one apply function.
- Distribution math is correct at the endpoints: Velocity spreads 8 to 127 across positions 1..16 with nearest-integer rounding, Pitch gives exactly one semitone per position at the default 0..15 range. Range setters clamp to descriptor bounds and cannot cross (min stays at or below max), covered by tests.
- Coexistence rules match the card: Full Level forces 127 pre-mapping, Pitch preserves played velocity so Full Level composes, Velocity owns velocity so Full Level reports unavailable while retaining its state for when 16 Levels turns off.
- Source lifecycle: enable uses last-played pad, Choose Source arms on the next press, target changes clear the source, and resolve syncs the target on every press so a retarget can never leak a mapped note from the previous target.
- Held-note release safety is preserved: ActiveInstrumentNote now carries the resolved velocity and the release still uses the stored track and pitch.
- The instrument logic moved into domains/perform/instrument.rs with its own test file, which also relieves the file-size pressure I flagged on #90. Also confirmed the #90 focus-loss fix landed as 8e1f1dd with its own regression file before merge; thanks for turning that around.
- Independently verified at this head: fmt clean, workspace clippy -D warnings clean, all 736 workspace tests pass, all eight CI jobs green. All touched files under 1,000 lines.

**One fix requested — a non-playable selection clears the 16 Levels source:**
sync_instrument_target is fed the raw project-wide selection at the update() entry, in actions.rs, and in both update_timeline.rs paths, with no playability filter. Card 12's contract says Audio Project Tracks never occupy an Instrument Target position, but this code assigns one internally: mid-performance, clicking an audio track row (Section overview, timeline) sets the internal target to that audio track and clears the 16 Levels source. Re-selecting the same MIDI track does not restore it, so the next pad press silently becomes a Choose Source press and plays the pad's unmapped pitch instead of the assigned level - exactly the kind of live surprise the lifecycle criteria exist to prevent. The lifecycle test only exercises two playable targets, so nothing covers this. Fix: treat non-playable selections as no target change (filter through is_playable_midi_target at the sync boundary so the last playable target is retained), plus a regression: enable 16 Levels with a source, select an audio track, re-select the original MIDI track, assert the source and mapping survive.

alexanderwanyoike:
Re-reviewed at aff3112. The fix request is resolved and then some: sync_instrument_target_from_selection ignores non-playable selections at every call site, and note resolution now keys off the retained instrument_target() instead of raw selection, so pads keep playing the last playable target even while an audio track is selected for editing - better than the minimum I asked for, and the exact regression I requested (selecting_a_non_playable_track_preserves_the_last_instrument_target_and_source) covers the audio round trip including the still-clears-on-real-retarget case. The control rail extraction into views_perform_instrument.rs (409 lines) keeps views_perform_pads.rs at 521.

Independently verified at this head: fmt clean, workspace clippy -D warnings clean, all 737 workspace tests pass, all eight CI jobs green. No open findings from my side; merge remains behind Alex's sign-off.

PR #92, opened 2026-07-20
Perform Card 14: add Note Repeat and Swing

alexanderwanyoike:
## Summary

- add engine-clocked Note Repeat with momentary N input, an on-screen latch, and eight straight/triplet rates
- compose Note Repeat with Card 13's resolved pitch/velocity boundary, including Full Level and 16 Levels
- place Full Level, 16 Levels, Note Repeat, Repeat Rate, and Track Swing in one compact right-side Instrument rail
- place project-wide Swing beside BPM in the shared transport, with per-Instrument-Target offsets and effective output feedback in the rail
- persist Project Swing and optional per-Project-Track offsets; generated straight-grid repeats swing while triplets and existing clip playback stay exact
- expose timestamped generated-note events for the later recording stack and document the scheduling boundary

## Review follow-ups

- pair momentary Note Repeat release with the runtime identity captured from the physical N press, including non-QWERTY layouts
- move snapshots and undo history into `state/snapshot.rs`; every touched Rust file remains below 1,000 lines
- group every direct-manipulation Swing drag into one undo step while keeping centered 1% nudge controls for exact correction
- prove repeats retain a 16 Levels-mapped velocity

## Verification

- `cargo test --workspace -j4` (complete workspace and doc-test pass; 317 vibez-ui tests and 158 engine tests)
- `cargo clippy -p vibez-ui --lib -j4 -- -D warnings`
- strict all-target Clippy reaches only pre-existing Rust 1.96 warnings in untouched VST3 host, engine test, and UI settings files
- isolated 1400×900 Xvfb visual/interaction QA with 16 Levels expanded, centered Swing icons, direct Project/Track Swing faders, exact 59%/75% mapping, and offset/effective feedback verified
- refreshed native dogfood demo from the stacked Card 14 worktree

## Stack

This PR targets `feat/card-13-full-level-16-levels` / PR 91 so Cards 13 and 14 land as the agreed single Instrument control surface. Recording workflows remain out of scope.

Do not merge without Alex's explicit approval.


alexanderwanyoike:
Reviewed at 7e7167d against Card 14 and the delivery contract.

**Verified working:**
- Engine-owned scheduling is done properly: an allocation-free fixed 16-voice array per track, per-frame sub-buffer triggering in both the playing and idle render paths, note-off-before-note-on retrigger consistent with the #85 ordering rule, and NoteRepeated events carrying sample-accurate timestamps for the later recording cards.
- The performance_position clock aligns with absolute transport position while playing and free-runs while stopped, and every clock-affecting change (play, stop, seek, BPM, project swing, track offset, sample rate) reschedules pending voices. Rate changes retime without retriggering, covered by rate_change_retimes_future_repeats_without_immediate_retrigger.
- Swing math is right: straight offbeats delayed by swing x interval/3 (2:1 triplet feel at 1.0), triplets exact at every swing amount, effective swing composed once at the EngineTrack boundary. SwingAmount/SwingOffset are clamped newtypes with NaN-rejecting deserialize.
- Persistence is complete and backward compatible: project-document round-trip with a legacy-default regression, save/load/new-project all resync the engine, and undo replay resends SetProjectSwing plus per-track SetTrackSwingOffset.
- Lifecycle is airtight where it matters: pad release stops that pad's repeats in both momentary and latched forms, WindowUnfocused stops repeating voices before the note-offs, and toggling repeat mid-hold starts/stops repeats for already-held notes. The rebind guard reserving N is a nice catch.
- The bit-identical existing-clip-playback regression (swing_does_not_change_existing_clip_playback) directly locks the card's central invariant.
- perform.rs went down from 991 to 897 lines via the input_mapping extraction. Independently verified at this head: fmt clean, workspace clippy -D warnings clean, all 740 workspace tests pass, all eight CI jobs green.

**Fix 1 — N momentary is layout-asymmetric and sticks on non-QWERTY layouts:**
The press matches physical Code::KeyN but the release matches logical Character "n". Per the iced 0.13 conversion established in the PR #72 investigation, KeyReleased carries only the logical key. On Colemak the physical N position types "k" and on Dvorak it types "b", so the release never matches is_note_repeat_release and momentary Note Repeat sticks ON after the key is released. The inverse hazard also exists: a pad rebound to a physical key whose logical character is "n" on the user's layout would have its release swallowed by the momentary branch, leaking the active_computer_keys entry and sticking a note (rebinding N itself is guarded, but other physical keys producing logical "n" are not). Fix: when the momentary press fires, record that press's runtime_key_id (the KeyPressed event carries both identities) and treat the release with the matching id as the momentary release, falling back to the current logical check. Please add a layout-mismatch regression alongside the existing keyboard tests.

**Fix 2 — state/mod.rs is at exactly 1,000 lines and this PR consumed the last line:**
Base was 999; the ProjectSnapshot field addition lands it exactly at 1,000. The contract's own wording in prior card evidence is that every changed file remains below 1,000, and the rule is split before crossing. Since this PR is the touch that used up the headroom, take the split here (ProjectSnapshot and its neighbors are a natural state/snapshot.rs cut).

**Fix 3 — swing edits mark the document dirty but are not undoable:**
SetProjectSwing and SetTrackSwingOffset are in marks_dirty and mutate project state, but no pre-edit snapshot is pushed, unlike the TrackMuteRequest path and the Section property edits they sit beside in that list. An undo after a swing tweak silently skips it. Either push a snapshot (at gesture start or on first change per drag to avoid history spam) or record the deliberate exception in the card so it is a decision rather than an accident.

**Integration note for the Card 13 merge (prominent so it survives conflict resolution):**
This branch is based on dev without #91, so StartNoteRepeat carries Card 12's plain pad pitch and fixed velocity. The card requires repeated notes to carry the velocity rules already in force, including Full Level and 16 Levels. Both PRs restructure domains/perform.rs, so the merge will conflict; whichever lands second must route resolve_instrument_note's resolved pitch and velocity into the repeat start (and ideally add a test that a 16 Levels velocity pad repeats at its mapped velocity, not the fixed one).

alexanderwanyoike:
Re-reviewed at 54d74be, now stacked on Card 13's branch with the MPC2000XL groove profile redesign.

**Fix 1 (N layout asymmetry) - resolved.** The momentary press records its own runtime key id and the release matches that id first with the logical-n fallback retained, exactly the suggested shape, with a non-QWERTY regression in the keyboard tests.

**Fix 2 (state/mod.rs at 1,000) - resolved.** Split into state/mod.rs (889) + state/snapshot.rs (122).

**Fix 3 (swing undo) - I was wrong, withdrawing it.** The router's centralized push in update.rs already snapshots before any dirty-marking Perform message, and both swing messages were in marks_dirty at the head I reviewed, so swing edits were already undoable. I compared against the mute path's local snapshot and missed the message-level mechanism. The new project_and_track_swing_edits_restore_through_undo restore coverage is welcome regardless, and the swing faders grouping drag gestures into one undo entry is the right call.

**Card 13 integration - resolved.** The press path routes resolve_instrument_note into the active note, and both ExternalNoteOn and StartNoteRepeat now carry the resolved pitch and velocity, with the promised repeated_sixteen_level_velocity_uses_the_resolved_pad_velocity test.

**MPC2000XL redesign - verified against the updated card.** GrooveProfile::Mpc2000XlV1 swings only 1/8 and 1/16; SwingAmount is the native 50..75% pair ratio with offsets in percentage points clamped to the profile range; offbeats quantize to the 96 PPQN tick grid like the hardware (a 1/16 pair spans 48 ticks, 66% rounds to tick 32 = exact triplet, 75% to tick 36 = 3:1), and the fixtures assert exactly those ticks. Legacy projects without groove data default to mpc_2000xl_v1 at straight 50%.

Independently verified at this head: fmt clean, workspace clippy -D warnings clean, all 758 workspace tests pass. Two administrative notes for merge sequencing: (1) CI has not run at 54d74be because the PR now targets the Card 13 branch - my local runs stand in for now, but once #91 merges and this retargets dev, wait for the fresh eight-job run before merging; the card's evidence is honest about this and about the dogfood capture predating the MPC correction, so the corrected feel still needs Alex's audition. (2) File-size headroom is thin in two touched files: project_format_v1.rs at 997 and mixer.rs at 991 - worth planning splits before Card 16 touches them. No other findings.

alexanderwanyoike:
Adversarial review of the Swing feel at 54d74be, prompted by Alex hearing it wrong. I hand-verified the low-level math first so effort isn't duplicated - these parts are CORRECT and not the cause: the 96 PPQN tick placement (66% on a 1/16 pair = tick 32 = exact triplet, 75% = tick 36 = 3:1), all positions resolve to integer ticks so there is no cumulative rounding drift, triplet grids are tick-exact, the fader emits ratio units end-to-end with 1% steps, and project+offset composition clamps correctly at the EngineTrack boundary.

**Root cause candidate 1 (high confidence): the swing grid is anchored to an arbitrary clock, so the audible pattern starts at a random phase - frequently REVERSED.**
next_repeat_sample derives steps as absolute multiples of the interval from sample 0 of performance_position's timeline. Two situations break the feel:

- **Transport stopped** (the natural way to audition Note Repeat): performance_position free-runs from wherever transport last stopped, accumulating callback frames forever. When you press a pad, the first repeat lands at an arbitrary distance, and - worse - the PARITY of the pair grid (which gap is long vs short) is arbitrary relative to your press. At 66% on 1/16 the correct pattern is long-short-long-short anchored on your press; about half the time you will hear short-long (anti-swing) or start mid-pair, and every new press re-rolls the dice. This exactly produces 'doesn't sound quite right'. On the MPC with the sequencer stopped, the first press IS the downbeat: the grid anchors to it.
- **Immediate Section launches**: the section's musical downbeat becomes the transition sample, but the swing grid stays on the absolute transport grid, so swing lands off the music's beat for the rest of that section. 1 Beat / 1 Bar launches are safe (their boundaries are even pair multiples).

Every engine test starts the engine and presses at position ~0, so grid and press are aligned by construction - the tests structurally cannot catch this.

**Suggested fix**: give the engine an explicit groove anchor sample. While playing, anchor = the musical grid as now (transport zero / section-aligned boundaries). While stopped, when the first repeat voice starts with no other voices active, set anchor = the press sample; steps become anchor + step x interval (+ swing displacement on odd steps), so the press is step 0, the first gap is the LONG one, and later pads phase-lock to the same anchor; clear the anchor when the last voice stops. Add fixtures that press mid-grid at a swung rate and assert the first two inter-onset intervals are long-then-short.

**Root cause candidate 2 (secondary, playing-transport feel): the unquantized source hit flams against the first scheduled repeat.**
The press sends ExternalNoteOn immediately AND schedules the first repeat strictly after the current position. Press slightly before a grid point and you get two hits milliseconds apart; with swing engaged the short gap makes this worse. On the MPC, while Note Repeat is held with the sequencer running, the press itself is quantized - the first audible hit is the next grid point. Suggested: while note repeat is active and transport is playing, suppress the immediate note-on and let the first scheduled trigger be the first sound (under the stopped-anchor model above, stopped behavior stays immediate because the press IS step 0).

Happy to re-verify whatever shape you and codex land on - the tick math itself can stay exactly as is.

alexanderwanyoike:
Re-reviewed the corrections delta at 0f71ff9. All six agreed items are implemented as decided: per-voice anchor samples with grid math relative to the anchor; the playing anchor derived from the active Section's start (fixing Immediate-launch phase) or transport zero in Arrange; the stopped anchor established by the first press, shared by later pads, and cleared when the last repeat stops; anchors preserved through swing/BPM/rate/offset changes; the immediate off-grid hit suppressed while repeating (the engine's first scheduled trigger is now the entrance); and the stopped-press extension documented as Vibez behavior, not MPC hardware behavior. The fixture suite covers stopped shared anchors, playing grid-wait entrances, and anchor preservation under swing changes. Verified within the stack: fmt clean, workspace clippy -D warnings clean, 789 tests pass at the stack top. Ready for the human audition pass.

PR #93, opened 2026-07-21
Perform Card 30: add live MIDI clip Swing

alexanderwanyoike:
## Summary

- persist an explicit Off / 1/8 / 1/16 Groove Grid on MIDI clips, defaulting legacy and new clips to Off
- map opted-in Arrange and resident Section playback through the MPC2000XL 96 PPQN pair shape without rewriting stored notes
- latch Swing at clip-local pair boundaries and guard shortened sources, same-pitch retriggers, Track offsets, and offline bounce parity
- place the compact Groove Grid selector in the shared clip detail strip

## Stack

- stacked on #92
- no merge requested

## Verification

- cargo test --workspace -j 4
- cargo clippy --workspace -j 4 -- -D warnings
- cargo fmt --all -- --check
- git diff --check
- off-screen native UI render at 1400x900

Exact fixtures cover 50%, 56%, 66%, and 75% ticks, Grid Off identity, off-grid monotonic mapping, pair-boundary latching, loop-local playback, prepared Sections, shortened boundaries, same-pitch note-off clamping, Track offsets, persistence, and bounce parity.

### DIFF
--- a/crates/vibez-core/src/midi.rs
+++ b/crates/vibez-core/src/midi.rs
@@ -1,6 +1,7 @@
 use serde::{Deserialize, Serialize};
 
 use crate::id::{ClipId, TrackId};
+use crate::perform::GrooveGrid;
 
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 pub struct MidiNote {
@@ -34,6 +35,8 @@ pub struct NoteClipInfo {
     pub loop_start_beats: f64,
     #[serde(default)]
     pub loop_end_beats: f64,
+    #[serde(default)]
+    pub groove_grid: GrooveGrid,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
@@ -79,6 +82,7 @@ impl TrackKind {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::perform::GrooveGrid;
 
     #[test]
     fn midi_note_frequency() {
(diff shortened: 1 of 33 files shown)


alexanderwanyoike:
Reviewed Card 30 at 1bc10fc. The implementation matches the adversarial design review point for point, and I probed the risky parts specifically:

- GrooveGrid::map_beat is the agreed monotonic piecewise-linear pair map with the breakpoint quantized to the 96 PPQN tick, so clips and Note Repeat land on identical samples; off-grid human timing maps without any tolerance heuristic and provably cannot reorder.
- The pair-start latch (per-clip Cell keyed by pair index) is sound under scrutiny: effective swing is constant within a buffer (commands drain at buffer start), pairs advance monotonically with the playhead, mapped positions cannot escape their pair because pair endpoints are fixed points, and wrap/entry forces a fresh latch. I could not construct a skip, double-fire, or reorder across buffer boundaries, mid-buffer Section splits, or swing decreases.
- The drop rule (mapped onset past a shortened boundary never fires) and the same-pitch note-off clamp are both at the shared crossed_beat boundary, and both render paths (Arrange and resident Section) go through the same helpers, so nothing is baked into prepared Sections and the knob is live mid-Section.
- GrooveGrid is its own enum rather than a reuse of NoteRepeatRate, persistence defaults legacy and new clips to Off with round-trip coverage, and the Cards 16/19 store-straight-plus-grid contract is recorded in the card.

One efficiency note, not blocking: mapped_note_end does a linear same-pitch scan per evaluation, which is O(notes^2) per buffer worst case for dense clips; fine at current clip sizes and allocation-free, but worth an index if profiling ever fingers it. One housekeeping note: engine_tests.rs (a legacy over-1,000 file) was touched with a 2-line edit; trivial touches of grandfathered files are reasonable, but a split card for the legacy six should exist eventually. Verified within the stack: fmt, clippy -D warnings, 789 tests at the top. No fix requests.

PR #94, opened 2026-07-21
Card 31: build Swing knobs and automation boundary

alexanderwanyoike:
## Summary
- replace only the Project and Target Swing faders with centered, prototype-aligned knobs
- support exact 50–75% numeric entry, 1% drag/wheel steps, fine drag, FOLLOW, contextual composition, and undo-grouped gestures
- add the stable persisted `track_swing_offset` automation target with AUTO precedence and pair-safe engine evaluation
- apply the automated effective Swing to Note Repeat, opted-in Arrange/Section clips, stopped audition, and offline rendering paths

## Stack
- Base: #93 (Card 30 live clip Swing)
- Depends transitively on #92 (Card 14 Note Repeat timing repair)
- Do not merge without Alex's explicit approval

## Verification
- `cargo test --workspace -q -j 4` (all pass; UI 321, engine 176)
- `cargo clippy --workspace -q -j 4 -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`
- offscreen visual + interaction pass at 1400×900: Project 59%, Target 63% / PROJECT +4, FOLLOW reset, centered pointers, no rail clipping

PR #95, opened 2026-07-21
Fix Arrange context menu dismissal race

alexanderwanyoike:
## What changed
- only install Arrange browser drag-hover handlers while a sample drag is active
- keep context menus open through passive Browser-domain and Dropbox catalog progress updates
- add regression coverage for both dismissal races

## Why
Right-clicking a clip rebuilt the Arrange view, which emitted a passive `DragHoverTrack` message. The global context-menu policy treated it—and an in-flight Dropbox catalog save—as unrelated actions and immediately cleared the menu.

## Verification
- `cargo clippy -p vibez-ui --lib -j 4 -- -D warnings`
- `cargo fmt --all --check`
- offscreen X11 interaction: right-click menu remains open after release and through catalog refresh

Stacked on #94 so the demo contains both requested fixes. No merge performed.

### DIFF
--- a/crates/vibez-ui/src/app/mod.rs
+++ b/crates/vibez-ui/src/app/mod.rs
@@ -146,6 +146,7 @@ mod project_replay;
 mod project_sections;
 mod update;
 mod update_media;
+mod update_policy;
 mod update_remote;
 mod update_timeline;
 mod views_arrangement;
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -51,6 +51,10 @@ fn context_menu_message_keeps_open(message: &Message) -> bool {
             | Message::View(ViewMsg::CursorMoved(_, _))
             | Message::View(ViewMsg::WindowResized(_, _))
             | Message::View(ViewMsg::MouseReleased)
+            | Message::KeyboardInput {
+                event: iced::keyboard::Event::ModifiersChanged(_),
+                ..
+            }
             | Message::Browser(_)
             | Message::RemoteCatalogPageFetched { .. }
             | Message::RemoteCatalogSaved { .. }
@@ -1095,5 +1099,13 @@ mod tests {
                 result: Ok(()),
             },
         ));
+    }
+
+    #[test]
+    fn modifier_state_sync_after_right_click_does_not_dismiss_context_menu() {
+        assert!(context_menu_message_keeps_open(&Message::KeyboardInput {
+            event: iced::keyboard::Event::ModifiersChanged(iced::keyboard::Modifiers::empty(),),
+            occurred_at: std::time::Instant::now(),
+        }));
     }
 }
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -27,6 +27,53 @@ fn apply_project_track_deletion_policy(message: Message, confirm: bool) -> Messa
     }
 }
 
+fn context_menu_message_keeps_open(message: &Message) -> bool {
+    matches!(
+        message,
+        Message::Tick
+            | Message::Transport(TransportMsg::EnginePosition(_))
+            | Message::EngineMetering { .. }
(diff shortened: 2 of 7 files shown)


alexanderwanyoike:
Live hardware follow-up fixed in `6651043`.

The synthetic replay missed a physical-input detail: Iced emits `KeyboardInput::ModifiersChanged(0)` immediately after the right-click. The blanket context-menu dismissal policy treated that modifier-state synchronization as a command and cleared both clip and empty-lane menus. The policy now preserves modifier sync events, with a regression copied from the live trace.

Also corrected the empty-lane coordinate contract so precise canvas clicks are no longer replaced by a stale global cursor position.

Verified with Alex on the real display: multiple clip and empty-lane menus remained open and only produced explicit dismissals. Temporary `[DEBUG-ctx91]` instrumentation has been removed. Full workspace tests pass (326 UI tests), UI lib Clippy passes with warnings denied, and formatting passes.

alexanderwanyoike:
Reviewed at 6651043. Two of the three changes are good fixes; the third deepens a fragile pattern and introduces a small regression. Two fixes requested.

**Good:**
- Gating the DragHoverTrack/ClearDragTarget handlers on an active browser drag is the correct root-cause fix: passive pointer movement over the timeline no longer emits Browser messages at all, so the original dismissal race dies at the source rather than being filtered downstream.
- Moving the ArrangementEmpty cursor-position fallback out of the view domain and into the three call sites is a genuine cleanup: the domain no longer special-cases one target by reaching into cursor state, callers explicitly supply the position, and the domain test now asserts the given coordinates are honored.

**Fix 1 - Message::Browser(_) is over-broad and adds a new UX bug.** The keep-open arm now covers every Browser message, including deliberate user actions: with an Arrange context menu open, clicking a Browser folder, selecting a sample, or starting an audition leaves the stale menu floating. Only the passive traffic needs keep-open (watcher reconciliation, catalog fetch/save progress, decode completions, drag-hover). Put the classification where it belongs: a BrowserMsg::keeps_context_menu() (or is_passive()) helper in the browser domain with an exhaustive match, so every future variant is forced to declare itself and the router just calls it. Same altitude argument as the strictness policy: update.rs should not own per-variant knowledge of another domain's messages.

**Fix 2 - update.rs was already over the 1,000-line limit (1073) and this PR grows it to 1111.** The contract is split before crossing, not grow-after-crossed. The extraction you already did points at the cut: move the context-menu dismissal policy (context_menu_message_keeps_open plus its tests, and the deletion-policy helper while you are there) into an app/context_menu.rs or app/update_policy.rs module and bring update.rs under 1,000.

**Debt to record, not a blocker here:** the keep-open allowlist design itself guarantees this class of bug recurs - every new passive message anywhere in the app dismisses open menus until someone remembers to add arm number 46. The structural fix is to stop inferring dismissal from message types entirely: render the context menu as an overlay with a backdrop that dismisses on outside press plus Escape, and delete the list. That is a small card, and I would rather see it soon than watch this list grow again; the same mechanism would serve the file menu.

Verified within the stack: fmt clean, workspace clippy -D warnings clean, 789 tests pass at the stack top. Note this PR targets the Card 31 branch, so CI has not run on it; it needs its own eight-job run once it retargets dev.

alexanderwanyoike:
Addressed both required review items in d16513c:\n\n- Browser menu behavior now lives behind an exhaustive BrowserMsg::keeps_context_menu() classification. Passive drag/watcher reconciliation traffic keeps menus open; deliberate Browser actions dismiss them. The regression covers both sides.\n- Router-level deletion and context-menu policies moved to app/update_policy.rs; app/update.rs is now 994 lines, below the 1000-line delivery limit.\n- The broader replacement of message-inferred dismissal with explicit backdrop/Escape/action lifecycle events is recorded separately as private Perform Card 32, outside this focused repair.\n\nVerification: cargo test --workspace -j 4; cargo clippy --workspace -j 4 -- -D warnings; cargo fmt --all -- --check; git diff --check. All pass.

PR #96, opened 2026-07-22
Deepen project transactions for recording

alexanderwanyoike:
## Summary

- add an explicit begin/commit/abandon project transaction around the existing Arc-shared snapshots
- keep selection and other runtime UI state outside canonical undo snapshots
- migrate Project Track deletion across Arrange and every Section onto one transaction
- document the recording transaction boundary in ARCHITECTURE.md

## Verification

- `cargo test -j 4 --workspace`
- `cargo clippy -j 4 --workspace -- -D warnings`
- focused transaction and project-domain tests
- native Xvfb dogfood: ordinary add/undo/redo and confirmed Project Track delete/undo/redo
- touched Rust line audit: maximum 996 lines

## Scope

Card 15 only. Section Record, Capture, and recording UI remain out of scope.

### DIFF
--- a/crates/vibez-ui/src/app/actions.rs
+++ b/crates/vibez-ui/src/app/actions.rs
@@ -807,15 +807,10 @@ mod perform_action_tests {
             arrange_timeline: Arc::clone(&state.arrangement.timeline),
             sections: Arc::clone(&state.perform.sections),
             bpm: state.transport.bpm,
-            bpm_text: state.transport.bpm_text.clone(),
             project_swing: state.perform.project_swing(),
             loop_enabled: state.transport.loop_enabled,
             loop_start_beats: state.transport.loop_start_beats,
             loop_end_beats: state.transport.loop_end_beats,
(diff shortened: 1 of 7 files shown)


alexanderwanyoike:
Reviewed Card 15 at 582c2da. Clean, no findings. The transaction is the minimal correct shape: begin captures one Arc-cheap snapshot plus the dirty flag, intermediate push_edit calls are swallowed while marking changed, commit pushes the before-snapshot as exactly one undo entry (empty transactions vanish), and abandon returns the pre-state with history untouched. Undo/Redo are guarded with a status message while a transaction is open. Selection, bpm_text, and clip/section selection are now excluded from snapshots entirely, satisfying the non-canonical-state criterion, and Project Track deletion is migrated as the first multi-store consumer with the stale-confirmation no-op correctly abandoning rather than committing an empty step. Verified: fmt clean, clippy -D warnings clean, 816 tests at the stack top, and this PR's own eight CI jobs are green. Ready for Alex.

PR #110, opened 2026-07-25
Perform Card 39: quantize Track Mute pad gestures

alexanderwanyoike:
## Summary

- add a global Immediate / 1 Beat / 1 Bar Track Mute timing preference, persisted in UI settings
- queue running beat/bar gestures on engine-owned sample boundaries, with repeated-gesture cancellation and stopped-transport immediacy
- materialize the effective change as one undoable project mutation and Capture event at the exact audible sample
- expose clear pending/effective pad state and document the engine/UI ownership protocol

## TDD evidence

- red: exact one-beat mute boundary had no quantization type or engine command
- green: sample-exact boundary split and effective engine event
- red/green cycles also covered stop cancellation, persisted default, domain request routing, deferred project mutation, one effective undo step, exact-grid ownership, and immediate-over-pending cancellation

## Verification

- `CARGO_TARGET_DIR=/home/alexander/Code/Apps/vibez-targets/card-15 nice -n 10 cargo test --workspace -j 4`
- `CARGO_TARGET_DIR=/home/alexander/Code/Apps/vibez-targets/card-15 nice -n 10 cargo clippy --workspace -j 4 -- -D warnings`
- `git diff --check` and the 1,000-line crossing audit
- isolated Xvfb producer pass: stopped immediate mute + undo; 1 Bar pending/effective; repeated-pad cancellation; preference persistence after restart

## Producer demo

1. Open Perform > Track Mutes and confirm the timing menu offers Immediate, 1 Beat, and 1 Bar.
2. While stopped, select 1 Bar and press a populated mute pad; confirm it changes immediately and Undo restores it.
3. Start transport, press the same pad, and confirm the pad shows a pending target and beat.
4. Let the boundary pass; confirm the pad becomes effectively muted exactly on the bar.
5. Queue the opposite change and press again before the boundary; confirm pending state clears and the current mute remains.
6. Restart Vibez and confirm the global timing preference is restored.

No merge requested.

### DIFF
--- a/crates/vibez-core/src/perform.rs
+++ b/crates/vibez-core/src/perform.rs
@@ -325,6 +325,43 @@ mod groove_tests {
     }
 }
 
+/// Reusable musical grid boundary for deferred Perform actions.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MusicalBoundary {
+    #[default]
+    Immediate,
+    OneBeat,
+    OneBar,
+}
+
+impl MusicalBoundary {
+    pub const ALL: [Self; 3] = [Self::Immediate, Self::OneBeat, Self::OneBar];
+
+    pub const fn label(self) -> &'static str {
+        match self {
+            Self::Immediate => "Immediate",
+            Self::OneBeat => "1 Beat",
+            Self::OneBar => "1 Bar",
+        }
+    }
+
+    /// Grid size for deferred boundaries. Immediate has no future grid.
+    pub const fn beats(self) -> Option<f64> {
+        match self {
+            Self::Immediate => None,
+            Self::OneBeat => Some(1.0),
+            Self::OneBar => Some(4.0),
+        }
+    }
+}
+
+impl std::fmt::Display for MusicalBoundary {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str(self.label())
+    }
+}
+
 /// Musical boundary at which a resident Section launch becomes effective.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
@@ -345,11 +382,19 @@ impl SectionLaunchQuantization {
     ];
 
     pub const fn label(self) -> &'static str {
+        if let Some(boundary) = self.musical_boundary() {
+            boundary.label()
+        } else {
+            "End of Section"
+        }
+    }
+
+    pub const fn musical_boundary(self) -> Option<MusicalBoundary> {
         match self {
-            Self::Immediate => "Immediate",
-            Self::OneBeat => "1 Beat",
-            Self::OneBar => "1 Bar",
-            Self::EndOfSection => "End of Section",
+            Self::Immediate => Some(MusicalBoundary::Immediate),
+            Self::OneBeat => Some(MusicalBoundary::OneBeat),
+            Self::OneBar => Some(MusicalBoundary::OneBar),
+            Self::EndOfSection => None,
         }
     }
 }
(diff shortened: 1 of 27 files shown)


alexanderwanyoike:
## Review: Card 39 quantized Track Mutes

The engine-owned boundary model is the right shape: the UI holds no timer, the
project track is only written when `TrackMuteChanged` comes back, and the
render split gives a genuinely sample-accurate transition. `queued_mute` is a
`Copy` field on `EngineTrack`, so the callback stays allocation-free, and the
segment loop cannot spin (`segment_frames >= 1` on every path).

One blocker, two fix requests, and some notes.

---

### 1. Blocker: a quantized mute is silently lost whenever its boundary lands at or past the arrangement loop end

`queue_track_mute` resolves the boundary with `next_grid_boundary`, which
returns the next absolute multiple of the grid on the transport clock
(`engine_section_queue.rs:279`). During Arrange playback with looping enabled,
`process_multitrack` splits at `loop_end` and resumes at `loop_start`
(`engine_render.rs:44-74`), so the position never reaches `loop_end` and never
advances past it. `apply_track_mutes_due` therefore never sees
`segment_start >= effective_at_samples`, and the split filter
(`*boundary < segment_end`) never fires either.

The mute never happens. The pad stays `PENDING` indefinitely, and nothing
clears it except Stop or a second tap on the pad.

With a bar-aligned one-bar loop and `1 Bar` quantization this fails **every
time**, because the next bar boundary is always the loop end. With a four-bar
loop, taps in the last bar stick. With `1 Beat`, taps in the last beat stick.
Looping a bar while playing with mutes is a completely ordinary thing to do.

Confirmed with a probe test against this branch (8 Hz sample rate, 120 bpm,
loop over samples 0..16, `OneBar` queue, 40 laps):

```
test review_probe_quantized_mute_at_the_arrangement_loop_end_never_becomes_effective ... FAILED
queued mute never became effective inside an arrangement loop
```

The Perform/Section path is unaffected because `performance_position` is
monotonic and never wraps (Card 35), so this is specific to the new use of the
grid boundary under an Arrange loop.

Please fix in the engine and add that regression. The boundary needs to be
resolved against the region the clock will actually traverse: either fold a
boundary at or beyond `loop_end` back onto the wrap point, or flush any pending
mute whose boundary was skipped when `process_multitrack` jumps to
`loop_start`. Treating the loop wrap as a musical boundary is also the
behavior a producer expects: the mute lands where the bar visibly restarts.

---

### 2. Fix: `TrackMuteQuantization` is a copy of `SectionLaunchQuantization`

`vibez-core/src/perform.rs:363-395` reproduces `SectionLaunchQuantization`
(`:339-361`) variant for variant, label for label, plus the same `ALL` and the
same `Display`. The beats mapping is copied too:

```rust
// engine_section_queue.rs:76-79
SectionLaunchQuantization::OneBeat => self.next_grid_boundary(now, 1.0),
SectionLaunchQuantization::OneBar  => self.next_grid_boundary(now, 4.0),

// engine_automation_commands.rs:161-165
TrackMuteQuantization::OneBeat => 1.0,
TrackMuteQuantization::OneBar  => 4.0,
```

Two enums now own the same musical fact and neither owns the beats. Please
extract one launch-boundary concept that answers `beats()` itself (Information
Expert), and let both Section launch and Track Mute name the subset they
accept. That is the descriptor-table point from the earlier reviews: the
variants are real now, so the knowledge belongs in one place, not spread across
two `match` arms in two crates.

While you are there: the `TrackMuteQuantization::Immediate => 0.0` arm in
`queue_track_mute` is dead. The `Immediate` case already returned above it.

---

### 3. Fix: the `ProjectSnapshot` literal is now hand-copied a fifth time

`engine_events.rs:19-28` rebuilds the snapshot field by field because it only
has `&mut AppState`. That literal already exists in `project_replay.rs:16`,
`undo_tests.rs:18`, `capture.rs:579` and `actions.rs:730`.

`take_snapshot` reads nothing but `self.state`. Move it to
`impl AppState { pub fn project_snapshot(&self) -> ProjectSnapshot }`, have
`App::take_snapshot` delegate, and collapse all five. The undo behaviour here
is correct (snapshot before mutation, `Arc::make_mut` protects it) so this is
purely about there being one definition of what a project snapshot is.

---

### Notes, your call

- **Pending pad label.** `PENDING MUTE @ BEAT 37.00` is an absolute beat on the
  engine clock. A producer cannot use that number mid-performance. A countdown
  ("IN 2 BEATS") or a bar position reads better, and it is shorter on a pad.
- **Cancellation status text.** `TrackMuteQueueCancelled` sets
  `"Track Mute change cancelled"`, but the engine also emits it for every
  pending track when the transport stops and when an immediate mixer mute
  supersedes a pending pad change. Stopping with three pending mutes writes that
  status three times. Consider setting the status only for a real second-gesture
  cancellation.
- **Meter events double on a split buffer.** `render_multitrack_frames` pushes
  one `TrackMeter` per track per call, so a buffer that splits at a boundary
  pushes two per track and the second wins. Harmless, but the reported peak for
  that block is the post-boundary segment only.
- **`perform_tests.rs` is at 998 lines** (996 on `dev`). Third time this epic a
  file has landed one or two lines under the limit. The new
  `perform_track_mute_tests.rs` was the right instinct; please also split
  `perform_tests.rs` before the next card touches it.

Everything else, including the Section-transition survival test, the stopped
immediacy path, the undo boundary, and the settings round-trip, reads correctly.

alexanderwanyoike:
Addressed the review in follow-up commit 1bb6372.

- The engine now treats an Arrange loop wrap as the crossed musical boundary and applies due queued Track Mutes before jumping to loop start. The regression test exercises a one-bar loop for repeated laps and proves the queued mute becomes effective at loop end.
- Section launch and Track Mute quantization now share one MusicalBoundary descriptor and beats mapping; the dead Immediate arm is gone.
- ProjectSnapshot construction now lives on AppState and all prior call sites delegate to it.
- Pending pads show a relative beat countdown, and cancellation copy is neutral for stop/supersession paths.

Verification: full workspace all-target tests pass; strict Clippy passes for the changed core/engine/UI libraries; CI run 30174908026 is green on Ubuntu, macOS, and Windows for Check and Test, plus Format and Clippy.

PR #111, opened 2026-07-25
Perform Card 40: paste clips at the focused playhead

alexanderwanyoike:
## Summary

- anchor the earliest copied Clip at the focused timeline playhead and preserve every musical and relative-Track offset
- give selected Section construction an independent runtime edit cursor, while Arrange retains its exact transport edit cursor
- preserve atomic compatibility checks, reminted identity, out-of-bound Section content, and one-undo-step paste behavior
- revise the architecture contract for focus-owned clipboard placement

## TDD evidence

- red: cross-timeline audio paste stayed at source sample 600 instead of focused sample 9000
- green: one earliest-beat anchor shifts audio and MIDI entries together after compatibility validation
- red: focus routing had no Section-owned playhead resolver
- green: focused Arrange/Section cursor resolution, selected-not-playing Section targeting, and pointer-seek isolation
- regression coverage includes multi-Track spacing, time-selection fragments, incompatible atomic failure, repeated reminting, out-of-bound Section content, and cut/paste Undo

## Verification

- `CARGO_TARGET_DIR=/home/alexander/Code/Apps/vibez-targets/card-15 nice -n 10 cargo test --workspace -j 4`
- `CARGO_TARGET_DIR=/home/alexander/Code/Apps/vibez-targets/card-15 nice -n 10 cargo clippy --workspace -j 4 -- -D warnings`
- `git diff --check`, architecture documentation, and the 1,000-line crossing audit
- isolated Xvfb producer pass with a freshly built binary; no physical desktop used

## Producer demo

1. In Arrange, create or select two compatible Clips on adjacent Tracks at different source positions.
2. Multi-select and copy them, click a later empty position on the first destination Track, and paste.
3. Confirm the earliest Clip begins at the white edit cursor and the second keeps both its musical and Track offset.
4. Undo and redo; confirm both pasted Clips leave and return together.
5. Copy the same phrase, open a selected Section, click its construction timeline, and paste.
6. Confirm the Section cursor—not Arrange or a playing Section—owns the anchor; scroll to verify later material remains stored at the Section edge.
7. Try an incompatible destination and confirm the visible failure leaves the timeline unchanged.

No merge requested.

### DIFF
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -85,6 +85,9 @@ impl App {
             // only computes the cross-domain context, routes the
             // message, and applies the returned action.
             Message::Transport(msg) => {
+                if self.place_focused_section_playhead(&msg) {
+                    return Task::none();
+                }
                 let stops_perform = matches!(&msg, crate::domains::transport::TransportMsg::Stop)
                     || matches!(
                         &msg,
@@ -140,14 +143,21 @@ impl App {
                 let clipboard_snapshot = msg
                     .is_clipboard_project_edit()
                     .then(|| self.take_snapshot());
+                let playhead_beats = self.focused_editor_playhead_beats();
+                let samples_per_beat = if self.state.transport.bpm > 0.0 {
+                    60.0 * self.state.transport.sample_rate as f64 / self.state.transport.bpm
+                } else {
+                    0.0
+                };
+                let playhead_samples = if self.focused_editor_is_section() {
+                    (playhead_beats * samples_per_beat).round().max(0.0) as u64
+                } else {
+                    self.state.transport.position_samples
+                };
                 let ctx = crate::domains::arrangement::ArrangementCtx {
-                    samples_per_beat: if self.state.transport.bpm > 0.0 {
-                        60.0 * self.state.transport.sample_rate as f64 / self.state.transport.bpm
-                    } else {
-                        0.0
-                    },
-                    playhead_samples: self.state.transport.position_samples,
-                    playhead_beats: self.state.position_beats(),
+                    samples_per_beat,
+                    playhead_samples,
+                    playhead_beats,
                 };
                 let action = self.route_arrangement_editor_message(msg, ctx);
                 if let (true, Some(snapshot)) = (action.mark_dirty, clipboard_snapshot) {
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -18,6 +18,19 @@ use super::*;
 
 impl App {
     pub(super) fn update(&mut self, message: Message) -> Task<Message> {
+        let message = match message {
+            Message::SectionTimeline(edit) => {
+                if super::update_timeline::section_timeline_claims_focus(
+                    self.state.view.workspace,
+                    self.state.perform.selected_section.is_some(),
+                ) {
+                    self.state.perform.editor_focus =
+                        crate::domains::perform::PerformEditorFocus::SectionConstruction;
+                }
+                *edit
+            }
+            message => message,
+        };
         let (message, undo_gesture) = match message {
             Message::UndoGesture { id, edit } => (*edit, Some(id)),
             message => (message, None),
@@ -67,6 +80,9 @@ impl App {
         }
 
         match message {
+            Message::SectionTimeline(_) => {
+                unreachable!("section timeline wrappers are removed before routing")
+            }
             Message::MenuItemSelected(overlay, action) => {
                 let task = self.update(*action);
                 menu_lifecycle::dismiss(&mut self.state, overlay);
--- a/crates/vibez-ui/src/app/update_timeline.rs
+++ b/crates/vibez-ui/src/app/update_timeline.rs
@@ -21,6 +21,10 @@ fn clipboard_targets_section(
         && focus == PerformEditorFocus::SectionConstruction
 }
 
+pub(super) fn section_timeline_claims_focus(workspace: Workspace, selected_section: bool) -> bool {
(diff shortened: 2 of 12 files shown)


alexanderwanyoike:
## Review: Card 40 paste at the focused playhead

The anchor arithmetic is right: earliest clip by `min_by(f64::total_cmp)`,
every other clip keeps its musical offset, track mapping untouched, and the
`playhead_samples`/`playhead_beats` pair in the context stays consistent for
both timelines. Deriving `playhead_samples` from the section cursor also fixes
split-at-playhead inside a Section editor, which previously used the Arrange
transport position (`domains/arrangement/mod.rs:844`) - a real latent bug fixed
for free, though nothing covers it yet.

Two fixes before this is done.

---

### 1. Undo resets the Section edit cursor to bar 1, so paste-after-undo lands in the wrong place

`playhead_beats` was added to `TimelineEditorState`, and every reload of a
Section editor rebuilds that struct from `Default`:

```rust
// domains/perform/sections.rs:30-36
pub fn load(&mut self, timeline: Arc<ArrangementTimeline>, selected_track: Option<TrackId>) {
    self.editor = TimelineEditorState { timeline, selected_track, ..TimelineEditorState::default() };
}
```

`sync_selected_section_editor` calls that, and it runs on every undo/redo
(`app/project_replay.rs:167`, inside `apply_snapshot`) and after a Section
Record commit (`app/section_record.rs:181`).

Repro:

1. Perform, select a Section, click the lane at bar 5 - cursor at beat 16.
2. Ctrl+V - clips land at beat 16. Correct.
3. Ctrl+Z.
4. Ctrl+V again - clips land at beat 0, and the cursor marker has visibly
   jumped back to bar 1.

That contradicts the acceptance line "Undo/redo and repeated paste are
deterministic", and it is the exact gesture a producer makes when a paste
lands wrong.

The fix is a placement decision, not a patch: the edit cursor is per-Section
runtime navigation state, not resolved-timeline state. It belongs next to the
timeline in `SectionTimelineEditor`, surviving reload the same way the Section
name and selection do not. PR #112 builds precisely that facility - a
`viewports: HashMap<SectionId, TimelineViewport>` on `SectionTimelineEditor`
that is stored and restored across `load_section` - and it deliberately keeps
zoom and scroll alive across the same reload path. The cursor should live with
it, not in `TimelineEditorState`.

That also removes a second smell: `TimelineEditorState::playhead_beats` is
never written for Arrange (Arrange reads `state.position_beats()` instead), so
today the field is dead for one of its two owners, and `update.rs` has to
branch on `focused_editor_is_section()` to decide which of the two meanings
applies.

---

### 2. Clicking a Section lane does not focus the Section editor, so the pointer workflow can silently seek Arrange instead

The card makes "clicking to place the playhead before pasting" the pointer
workflow, and `place_focused_section_playhead` only intercepts `SeekToBeat`
when `editor_focus == SectionConstruction`. But the focus message never fires
for a lane click:

- The Section pane is wrapped in `mouse_area(construction).on_press(FocusEditor(SectionConstruction))`
  (`views_perform_sections.rs:562`).
- iced's `mouse_area` returns early when its child captures the press
  (`iced_widget-0.13.4/src/mouse_area.rs:211-221`).
- `TrackClipCanvas` captures on `ButtonPressed` before it ever emits anything
  (`widgets/timeline/clips.rs:451-456`).

So focus is set only by `select_section` (`domains/perform.rs:910`) and by
clicking dead space. The pad surface has the mirror-image problem: pads are
`mouse_area` themselves, so only a click in the gaps between pads sets
`PadSurface`.

Repro:

1. Select a Section - focus becomes `SectionConstruction`.
2. Click a gap in the pad surface - focus becomes `PadSurface`.
3. Click in the Section lane to place the cursor.
4. `SeekToBeat` is not intercepted: the Arrange transport moves and
   `EngineCommand::Seek` goes to the engine, mid-performance.

The seek-the-arrangement half is pre-existing, but this card is what makes the
lane click a deliberate, documented gesture, so it should own the fix. Setting
focus explicitly from the Section lane and ruler press (rather than relying on
a wrapper that the canvas always swallows) is the smallest honest fix. This
matters more once #112 lands, because that PR adds a Section ruler whose entire
purpose is to be clicked.

---

### Observation, your call

Arrange has no edit cursor separate from the transport, so
`focused_editor_playhead_beats` returns `state.position_beats()` and a paste
while the arrangement is playing lands at wherever playback happens to be at
that instant - a different bar every time. Ableton pastes at a stationary
insert marker. The Section side of this PR now has the better behaviour of the
two: an independent cursor that playback does not move. Worth deciding whether
Arrange should grow the same thing, or whether stop-then-paste is good enough
for the release.

---

### Merge ordering

This PR and #112 both replace the `-1.0` playhead argument to
`TrackClipCanvas::from_track` in `views_perform_sections.rs`: this one with the
edit cursor, #112 with the playing-Section position, and #112 also deletes the
separate overlay playhead line. A textual merge will quietly pick one and drop
the other. The Section editor needs both markers - an edit cursor and a
playback playhead - so please resolve that deliberately rather than by
rebase. See the note on #112.

alexanderwanyoike:
Addressed the review in follow-up commit 3de7da1.

- The Section edit cursor is now per-Section runtime navigation state owned by SectionTimelineEditor, so reload, undo/redo, Section Record synchronization, and switching Sections preserve it.
- Repeated paste after undo is covered and remains anchored at the preserved cursor.
- Section ruler and lane canvas messages are explicitly wrapped so those child canvases focus Section construction before routing the click; they no longer rely on a parent mouse area that can be swallowed.
- TimelineEditorState no longer carries a Section-only cursor.

Verification: a cold full workspace all-target test run passes; strict Clippy passes for the changed UI library; CI run 30175369135 is green on Ubuntu, macOS, and Windows for Check and Test, plus Format and Clippy.

PR #112, opened 2026-07-25
Perform Card 41: bring Section navigation to Arrange parity

alexanderwanyoike:
## Outcome

Brings Section construction navigation to Arrange parity without sharing
viewport state with Arrange or another Section:

- continuous pointer-anchored zoom, horizontal wheel/trackpad pan,
  middle-drag pan, zoom-to-fit, and last-full-view clamping
- an Arrange-style ruler and minimap fixed above the vertically scrolling
  Track body
- one viewport geometry for ruler, minimap, lanes, automation, recording
  preview, playhead, drag/drop, and bounded materialization
- a smooth quadratic 64 px edge zone capped at 8 beats/second
- drag compensation for viewport movement, with release/zone exit stopping
  edge-scroll immediately

Each Section keeps an independent runtime-only viewport. Project bytes and
Arrange/Section content ownership are unchanged.

## Diagnosis

The previous behavior had two competing scroll paths:

1. action-driven auto-scroll used a hard-coded 1400 px Arrange canvas and
   jumped on every drag update;
2. the 60 fps path used hard-coded 1600/230 px edges and could advance up to
   120 beats/second.

Section drags set the Section editor's drag flag, so they missed the tick path,
while action routing could still mutate the unrelated Arrange viewport. A move
drag then reinterpreted its pointer through the changed viewport. This PR gives
the focused editor one explicit viewport and one bounded tick policy.

## TDD evidence

Red tests were added first for:

- edge-zone velocity and the conservative maximum
- move-drag compensation after viewport motion
- pointer-anchored zoom and last-full-view clamping
- independent viewport restoration for each Section
- sticky navigation-header structure
- playhead/lane geometry alignment
- viewport-bounded long-Section materialization

All are green in the final suite.

## Validation

- `cargo test -p vibez-ui -j 4` — 401 passed
- `cargo test --workspace -j 4` — passed, 0 failed
- `cargo check --workspace -j 4` — passed
- `cargo clippy --workspace -j 4 -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed
- architecture documentation updated
- no new Rust file crossed the repository's 1,000-line threshold
- isolated Xvfb dogfood used the current binary; no physical desktop was used
- [cross-platform CI run 30172419488](https://github.com/alexanderwanyoike/vibez/actions/runs/30172419488)
  — all eight Check, Test, Clippy, and Format jobs passed on Ubuntu, macOS, and
  Windows

Xvfb evidence is retained at
`/home/alexander/Code/Apps/vibez-demo-artifacts/card-41-section-navigation/`.
It covers a 12-bar Section, ten Tracks, sticky vertical scrolling,
pointer-anchored zoom, pan in both directions, per-Section viewport restoration,
navigation during playback, and a held edge resize.

## Producer demo checklist

1. Create at least ten Project Tracks, enter Perform, and create Section 01.
2. Extend Section 01 to 12 bars.
3. Hover a lane and Shift-scroll around a recognisable bar; confirm that bar
   stays under the pointer while ruler, minimap, and lanes zoom together.
4. Plain-scroll horizontally in both directions, then middle-drag the lane;
   confirm smooth pan and stable end clamping.
5. Scroll vertically from the Track gutter to the last Track; confirm the
   ruler and minimap remain visible and horizontally aligned.
6. Create Section 02, change its viewport, switch between Sections 01 and 02,
   and confirm each restores its own zoom and pan.
7. Launch Section 01 and repeatedly zoom, pan, and vertically scroll; confirm
   playback and the local Section playhead keep advancing without UI stalls.
8. Add a MIDI Clip, zoom in, and drag its move or right-edge handle into both
   viewport edge zones. Hold briefly, leave the zone, and release. Confirm
   scrolling ramps gently, stops immediately, and the edit lands under the
   pointer without extra bars.

## Delivery isolation

This PR targets `dev` directly and contains one Card 41 commit based on
`045ae1ebe6b05eb213a40d09963985276547806a`. It does not contain Card 39 or
Card 40.

### DIFF
--- a/crates/vibez-ui/src/app/actions.rs
+++ b/crates/vibez-ui/src/app/actions.rs
@@ -13,6 +13,9 @@ use crate::state::{ArrangementSelection, DetailPanelTab, UiEffect};
 
 use super::*;
 
+/// Keep edge-scroll activation inside the window border/shadow hit area.
+const TIMELINE_WINDOW_EDGE_INSET: f32 = 8.0;
+
 fn apply_track_mute_request(
     project_tracks: &mut Arc<crate::state::ProjectTracksState>,
     history: &mut crate::state::UndoHistory,
@@ -686,7 +689,7 @@ impl App {
         // A stationary pointer does not produce canvas events. Drive the
         // active editor's bounded edge-scroll from the 60fps UI clock.
         let cursor_x = self.state.view.cursor_x;
-        let right_boundary = (self.state.view.window_width - 8.0).max(1.0);
+        let right_boundary = (self.state.view.window_width - TIMELINE_WINDOW_EDGE_INSET).max(1.0);
         let section_drag = self.state.view.workspace == crate::state::Workspace::Perform
             && self
                 .state
--- a/crates/vibez-ui/src/app/actions.rs
+++ b/crates/vibez-ui/src/app/actions.rs
@@ -177,7 +177,16 @@ impl App {
             self.state.view.detail_panel_tab = DetailPanelTab::Clip;
         }
         if let Some(beat) = action.scroll_to_beat {
-            self.auto_scroll_to_beat(beat);
+            if !self.state.arrangement.drag_resize_active
+                && !self
+                    .state
+                    .perform
+                    .section_editor
+                    .editor()
+                    .drag_resize_active
+            {
+                self.auto_scroll_to_beat(beat);
+            }
         }
         if let Some((start, end)) = action.loop_from_selection {
             let mut engine = crate::domains::EngineTx(&mut self.cmd_tx);
@@ -241,6 +250,11 @@ impl App {
         }
         if action.end_drag_resize {
             self.state.arrangement.drag_resize_active = false;
+            self.state
+                .perform
+                .section_editor
+                .editor_mut()
+                .drag_resize_active = false;
         }
         if action.close_device_menu {
             self.state.devices.context_menu = None;
@@ -359,7 +373,16 @@ impl App {
             }
         }
         if let Some(beat) = action.scroll_to_beat {
-            self.auto_scroll_to_beat(beat);
+            if !self.state.arrangement.drag_resize_active
+                && !self
+                    .state
+                    .perform
+                    .section_editor
+                    .editor()
+                    .drag_resize_active
+            {
+                self.auto_scroll_to_beat(beat);
+            }
         }
         if action.drag_resize_active {
             if self.state.view.workspace == crate::state::Workspace::Perform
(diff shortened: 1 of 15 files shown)


alexanderwanyoike:
## Review: Card 41 Section editor navigation

This is the strongest of the three. Replacing the fixed 160px-per-bar canvas
inside a horizontal scrollable with a real `TimelineViewport` puts the Section
editor on the same geometry as Arrange, and the per-Section viewport map is
exactly the right ownership: runtime navigation state that survives a reload
because it is not part of the resolved timeline. `edge_scroll_velocity` and
`compensated_drag_beat` are pure, bounded, and actually tested. The diagnosis
in the card is also confirmed by the code: the old tick path used a hard-coded
1600/230px Arrange geometry and could move 120 beats per second.

Two things worth checking before this lands: resize and loop drags do **not**
need the new compensation, because they resolve through
`self.x_to_beat(local_x)` on the live geometry (`clips.rs:658`), so they were
already viewport-stable; only `Move` carried a stored start pixel and it is the
one that got compensated. And the ruler/lane alignment holds despite the
vertical scrollable reserving 6px on the right
(`iced_widget-0.13.4/src/scrollable.rs:422-436`), because both surfaces map
beats through the same zoom/scroll geometry rather than through their widths.

---

### 1. Do not merge this before #111: the new ruler seeks the arrangement

`RulerWidget` emits `Message::Transport(TransportMsg::SeekToBeat(beat))` on a
short click (`widgets/timeline/ruler.rs:439`). Nothing in this PR intercepts
it, so clicking the brand-new Section ruler moves the Arrange transport and
sends `EngineCommand::Seek` to the engine - during a performance.

The lane canvas already had this behaviour, so it is not newly introduced, but
this PR adds a ruler whose whole purpose is to be clicked, which turns a latent
wart into the obvious gesture. #111 adds exactly the interception that fixes it
(`place_focused_section_playhead`), subject to the focus problem I raised
there. Please land #111 first and rebase this on it.

The other ruler and minimap messages are fine:
`ViewMsg::ScrollArrangement` and `ViewMsg::ZoomAround` are picked up by
`route_section_timeline_navigation`, and `ArrangementMsg::SetTimeSelection`
routes through the focused-editor resolver. The ruler has no loop-drag message,
so passing `loop_start_beats: 0.0, loop_end_beats: total_beats` is display only.

### 2. Semantic conflict with #111 on the Section playhead

Both PRs replace the same `-1.0` argument to `TrackClipCanvas::from_track`:
this one with the playing-Section position, #111 with the edit cursor. This PR
additionally deletes the `stack!`-based `section_playhead_line` overlay, so
whichever way the merge resolves, one of the two markers disappears silently.
The Section editor wants both. Please decide it explicitly rather than letting
the rebase pick.

### 3. Three of the new tests assert nothing

```rust
let ruler_x = viewport.geometry().beat_to_x(beat);
let lane_x  = viewport.geometry().beat_to_x(beat);
assert_eq!(ruler_x, lane_x);
```

`section_playhead_and_lanes_share_the_same_viewport_alignment` compares an
expression with itself. `ruler_and_minimap_form_one_sticky_header_above_the_track_scroller`
asserts `A + B == A + B`. `long_section_materialization_stays_bounded_to_the_visible_viewport`
asserts `x.max(1.0) == x` for `x = 824.0`.

The card claims coverage of "playhead alignment" and "sticky header layout";
these three provide none of it. Either make them assert something that can
fail - the ruler and the lane resolving the same beat through *independently
constructed* geometry from the shared viewport would be a real test - or drop
them rather than bank coverage that is not there.

### 4. Please re-test scrolled Section clip rendering by hand

This rewrites the exact code path of two bugs you found last week -
`932075b` "render clips across scrolled sections" and `b4de3f2` "keep split
MIDI fragments visible" - and it deletes the test that pinned the fix
(`horizontally_scrolled_section_materializes_its_full_timeline`, whose comment
was literally "late split fragments must exist in the canvas before the parent
scroll reveals them"), replacing it with the tautology above.

The new model is correct in principle: culling is now scroll-aware with a 25%
prefetch window (`clips.rs:141-146`), which is the proven Arrange behaviour, and
a viewport-bounded canvas is the right answer once the canvas no longer spans
the whole timeline. But the guarantee is now untested, so a long Section,
scrolled and zoomed, with split MIDI fragments late in the timeline, deserves a
hands-on look before sign-off.

---

### Notes, your call

- **Magic numbers.** `section_timeline_viewport_width` subtracts a bare `8.0`
  and the tick handler uses `window_width - 8.0`. The real reservation is 6px
  (scrollbar width 5 + spacing 1). Deriving it from the scrollbar constants
  keeps it honest if that styling ever changes.
- **Arrange shortcuts while a Section is focused.** `route_section_timeline_navigation`
  returns early, so Zoom In/Out/To Fit and scroll no longer reach Arrange while
  a Section has focus. That is intended, but combined with the focus problem in
  #111 it means a stray click in the pad-surface gaps sends those shortcuts to
  Arrange instead, invisibly.
- **`remove_viewport` on delete** is handled; the map is bounded by Section
  count. Good.

alexanderwanyoike:
Addressed the review in follow-up commit d2c350b while keeping this PR unstacked against the approved dev base.

- Diagnosed the ruler path end to end: RulerWidget emitted a generic Transport seek, which the app routed to Arrange because edit and playback positions were conflated. Section ruler/lane/minimap messages now explicitly focus and route through the selected Section timeline, and Section seeks update its per-Section edit cursor without moving Arrange or sending an engine seek.
- The Section editor now owns both independent truths: a persistent per-Section edit cursor and a separate engine-owned playback marker. The lane renders both simultaneously.
- Removed the three tautological tests. Replacements construct ruler and lane geometry independently, exercise actual viewport culling of two late split MIDI fragments in a 128-beat Section, preserve independent cursors across Section switches, and cover Section seek routing.
- Re-tested a long scrolled and zoomed Section in isolated Xvfb: late split MIDI material remained visible, sticky ruler/minimap alignment held, live playback showed distinct edit/playback markers, and returning to Arrange confirmed its cursor remained at beat zero. No physical desktop was used.
- The 6 px scrollbar reservation is derived from named width/spacing constants; the remaining window-edge inset is named and documented.

Verification: all 402 UI tests and the full workspace all-target suite pass; strict Clippy passes for the changed UI library; CI run 30176108956 is green on Ubuntu, macOS, and Windows for Check and Test, plus Format and Clippy.

PR #119, opened 2026-07-27
fix(ui): use the platform command modifier for every shortcut

alexanderwanyoike:
On macOS the app used two different modifiers. `Cmd+1` narrowed the grid, but copy was `Ctrl+C` and undo was `Ctrl+Z`. One app, two conventions, and the wrong one for the platform.

## Cause

`keyboard.rs` gated the grid shortcuts on `modifiers.command()`, which iced resolves to **Cmd on macOS and Ctrl elsewhere**. Everything else — clipboard, undo/redo, add track, split, join, loop, zoom — was gated on `modifiers.control()`, the **literal Control key on every platform**. The clip lane canvas (`Ctrl+D/E/J`) had the same problem.

## Fix

Both now resolve through one helper:

```rust
pub(crate) fn command_held(modifiers: Modifiers, on_macos: bool) -> bool {
    if on_macos { modifiers.logo() } else { modifiers.control() }
}
```

Passing the platform as an **argument** rather than reading `cfg!` inside the body is the point of the design. `Modifiers::command()` resolves at compile time, so on Linux it silently answers Ctrl and a macOS regression is invisible to a Linux test run. As a parameter, both platforms are asserted from any host.

The piano roll's select-all already used `command()` and moves to the helper too, so there is a single policy in the tree rather than three spellings of the same idea.

## Test-first

Worth being straight about the red/green here. The first test I wrote used the platform modifier through `global_key_handler`, and it **passed before the fix on Linux**, because `command()` is Ctrl on this host. That is not a test, it is a tautology on one platform.

So the seam came first, and its test is a genuine red — it failed to compile because `command_held` did not exist, then passed once it did:

```rust
assert!(command_held(Modifiers::LOGO, true));
assert!(!command_held(Modifiers::CTRL, true));
assert!(command_held(Modifiers::CTRL, false));
assert!(!command_held(Modifiers::LOGO, false));
```

Added alongside it:
- `editing_shortcuts_share_the_grid_shortcuts_modifier` — the same modifier reaches grid, paste, undo and zoom.
- `the_super_key_never_stands_in_for_the_command_modifier_off_macos` — guards the lazy fix of accepting both Ctrl and Logo everywhere, which would capture the desktop's own Super shortcuts on Linux and Windows.

**Two existing tests hard-coded `Modifiers::CTRL`** for command shortcuts (`create_clip_shortcut_accepts_shifted_m` and the track-reorder assertion). They pass today and would have started failing on macOS CI the moment this landed. Both now use the platform modifier the production code uses.

## Behaviour change

**macOS only.** `command_held(.., false)` is `modifiers.control()`, exactly what Linux and Windows did before, so those platforms are byte-identical. On macOS, command shortcuts move from Ctrl to Cmd, which is what a Mac user expects.

## Site

The landing page documented the split as a known-unfixed quirk. That paragraph is gone. The tables said `Cmd` for grid rows and `Ctrl` for everything else; they now say `Ctrl` throughout with a single note that macOS uses Cmd.

This PR also carries the merge of `main` into `dev`. The site went to `main` directly, which skipped the normal feature→dev→main flow and left `dev` without `site/` or the Pages workflow. Merging here resyncs the two.

## Checks

`cargo test -p vibez-ui --lib` 416 passed, `cargo clippy --workspace --all-targets --exclude vibez-plugin-host -- -D warnings` clean, `cargo fmt --all -- --check` clean. Linux only, so macOS CI on this PR is the first real exercise of the new behaviour.


alexanderwanyoike:
## Review: engineering consistency and code quality

Ran the checks locally on this branch: 416 vibez-ui tests pass, clippy clean with `-D warnings`.

### What's good

- **The seam design is right.** Passing the platform as a parameter instead of reading `cfg!` inside `command_held` is what makes both platforms assertable from one host, and the PR body is honest that the first test attempt was a tautology. The test helper `command_modifier()` re-derives the platform mapping independently via `#[cfg]`, so it is a genuine oracle rather than an echo of the production code.
- **The negative test is the valuable one.** `the_super_key_never_stands_in_for_the_command_modifier_off_macos` blocks the lazy "accept both" fix that would eat the desktop's Super shortcuts, and it asserts in both directions per platform.
- **Latent CI failures fixed preemptively.** The two existing tests hard-coding `Modifiers::CTRL` would have broken macOS CI the moment this landed. CI runs `macos-latest`, so the `cfg(target_os = "macos")` branches genuinely execute.
- Docs updated in the same change: the site tables now say Ctrl with one Cmd note, matching the new behavior.

### Findings

**1. The bare-key guards still exclude literal Ctrl, not the command modifier** (`keyboard.rs:379` and `:390`). The Delete/Backspace guard is `!modifiers.control()` and the B draw-mode guard is `!modifiers.control() && !modifiers.shift()`. On macOS after this PR those guards check the wrong key: Cmd+B toggles draw mode and Cmd+Delete fires delete, while Ctrl+B and Ctrl+Delete are the combinations that get blocked, the inverse of the new policy. The Enter and Alt+arrow guards nearby already exclude both `control()` and `logo()`, so this is an internal inconsistency within the same function. Suggest `!command_held(modifiers, ON_MACOS)` (or excluding both keys, as the Enter guard does) so "no command modifier held" means the same thing everywhere. This is the same class of bug the PR exists to kill, so worth fixing before merge.

**2. One modifier check remains untouched: `automation_lane.rs:334`** tracks `m.control()` for the sweep-erase drag and its crosshair cursor. It is a mouse-gesture modifier rather than a keyboard shortcut, so it may be a deliberate carve-out, but on macOS Ctrl+click is the right-click convention, and the site now promises "a single command modifier per platform". Either move it to `command_held` or leave a comment saying why erase-drag is intentionally literal Ctrl; silence will read as a missed site to the next person who greps.

**3. Minor duplication in tests.** The pre-existing `command_number_shortcuts_control_the_shared_grid` test re-implements the platform pick inline (`#[cfg] let command = Modifiers::LOGO / CTRL`) directly below the new `command_modifier()` helper that expresses the same thing. Worth folding into the helper while here.

**4. Canvas call sites have no direct coverage.** The clip-lane D/E/J and piano-roll Cmd+A changes are covered only transitively through the `command_held` unit test. `canvas::Program::update` is awkward to drive in a test, so this is acceptable, but those two sites depend on the helper being wired in correctly rather than on any assertion.

### Verdict

The core fix is solid and well-tested for what a unit test can reach. Finding 1 is the one to fix before merge; 2 and 3 are small follow-ups. The site and workflow content riding along is fine in quality (self-contained, accessible, sensible Pages concurrency settings), though bundling the resync merge with a behavior fix makes the diff read larger than the change is.

alexanderwanyoike:
Thanks, that's a sharp review. Findings 1 and 2 were both real and both fixed in 16fb52a; 3 done; 4 acknowledged as-is.

## 1. Bare-key guards — fixed, and worse than described

Confirmed by reading it back. `Delete`/`Backspace` was `!modifiers.control()` and `B` was `!modifiers.control() && !modifiers.shift()`. On macOS after the first commit those were exactly inverted: **Cmd+B toggled draw mode and Cmd+Delete deleted, while Ctrl+B and Ctrl+Delete were the blocked pair.**

What makes it worse than an oversight is that the `Enter` guard eight lines above already excluded `control`, `alt`, `shift` **and** `logo`, so the file disagreed with itself inside one screen. Both now ask `command_held(modifiers, ON_MACOS)`, so "no command modifier held" means one thing everywhere.

Added `bare_key_shortcuts_are_blocked_by_the_command_modifier_not_literal_control`, asserting bare `B` and bare `Delete` fire while the command-modified pair does not. Same caveat as the rest: green on Linux either way, genuinely exercised on macOS CI.

## 2. `automation_lane.rs` — converted, not documented

I went looking for the deliberate carve-out and could not justify one. `state.ctrl` gates a **Ctrl + left-click-drag** sweep-erase plus its crosshair cursor. Literal Ctrl is a bad choice there for a reason beyond consistency: **Ctrl+left-click is the macOS right-click gesture**, so the erase drag was colliding with the context menu on the one platform where it was already wrong.

It now shares `command_held`, and I renamed the field `ctrl` to `command`, since a bool called `ctrl` holding Cmd on macOS is the next person's confusion.

## 3. Test duplication — folded

`command_number_shortcuts_control_the_shared_grid` now calls `command_modifier()`. Worth noting for anyone reading later: the helper still derives the mapping from `#[cfg]` independently rather than calling `command_held`, so it stays an oracle and does not become an echo of the code under test.

## 4. Canvas coverage — agreed, left alone

Accurate. The clip-lane and piano-roll sites lean on the helper being wired correctly rather than on an assertion. `canvas::Program::update` needs a `Renderer` and real bounds to drive, which is a bigger lift than this fix warrants. The shortcut catalogue plus conformance test is the right home for that.

## On the bundled merge

Fair hit. Bundling the `main`→`dev` resync with a behaviour fix makes the diff read much larger than the change is. That resync only exists because I merged the site PR straight to `main` and skipped `dev`, so this is me cleaning up after my own shortcut. Worth separating if you want a clean diff to review, though splitting now costs a rebase.

`cargo test -p vibez-ui --lib` 417 passed, clippy clean over `--all-targets`, fmt clean.

PR #121, opened 2026-07-28
fix(browser): name the Dropbox connection after the provider, not a person

alexanderwanyoike:
Found by @alexanderwanyoike opening v0.1.0 on a different Mac: the Browser's Remote section showed **"Alex's Dropbox"** on a machine with no connection to that account.

## It is hardcoded

`crates/vibez-ui/src/remote_provider.rs`

```rust
pub const DROPBOX_CONNECTION_NAME: &str = "Alex's Dropbox";
```

Nothing derives it from the signed-in account. **Every user on every platform sees it**, and it shipped in v0.1.0.

## It is not only cosmetic

The same constant is written into saved projects as media provenance for remote clips (`async_helpers.rs:344`), so it lands in other people's project files too, not just their UI.

## Fix

- The constant now reads `"Dropbox"`.
- `views_browser.rs` had its **own copy of the literal** as a fallback rather than using the constant. That duplication is how a single rename could have missed a user-visible site, so it now references the constant.
- The other three occurrences were fixture strings in project-format round-trip tests and the format-proof binary. Arbitrary as far as those tests care, but they carried the same personal name, so they move too.

## Guard

```rust
assert_eq!(DROPBOX_CONNECTION_NAME, "Dropbox");
assert!(!DROPBOX_CONNECTION_NAME.contains('\''),
    "a possessive apostrophe suggests a personal name crept back in");
```

Tautological on the first line by design: it is a tripwire, and the comment records that this shipped personalised in v0.1.0.

## Also checked

Swept the tree for other personal identifiers in shipped code and assets. The only other hit is `io.github.alexanderwanyoike.vibez` in the macOS `Info.plist`, which is correct reverse-DNS from the GitHub org and should stay.

## Not done here

Showing the **actual connected account name** would be better UX than a generic label. That needs a Dropbox account-info call, so it is a feature rather than this fix.

## Checks

Full workspace tests pass (the project-format round-trips exercise the changed fixtures), clippy clean over `--all-targets`, fmt clean.

Worth landing before the Show HN post.

### DIFF
--- a/crates/vibez-project/src/bin/project-format-v1-proof.rs
+++ b/crates/vibez-project/src/bin/project-format-v1-proof.rs
@@ -60,7 +60,7 @@ fn run_measurement(output_dir: &Path) -> Result<(), Box<dyn Error>> {
             provenance: Provenance::Remote {
                 provider: "dropbox".into(),
                 connection_id: "alex-dropbox".into(),
-                connection_name: Some("Alex's Dropbox".into()),
+                connection_name: Some("Dropbox".into()),
                 source_id: "id:proof-remote-vocal".into(),
                 source_path: "/Megalodon/Vocals/remote-vocal.wav".into(),
                 revision: Some("proof-rev-1".into()),
--- a/crates/vibez-project/tests/project_format_v1.rs
+++ b/crates/vibez-project/tests/project_format_v1.rs
@@ -345,7 +345,7 @@ fn remote_import_becomes_self_contained_project_media_with_safe_provenance() {
         vibez_core::track::MediaProvenance::Remote {
             provider: "dropbox".into(),
             connection_id: "dropbox-primary".into(),
-            connection_name: Some("Alex's Dropbox".into()),
+            connection_name: Some("Dropbox".into()),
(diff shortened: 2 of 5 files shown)


alexanderwanyoike:
Agreed, dropped it in ebfb8b4.

`assert_eq!(DROPBOX_CONNECTION_NAME, "Dropbox")` restated the constant. It fails only when the value changes, which git already reports, and it exercised no behaviour through any interface. Textbook change detector.

The apostrophe assertion looked like it encoded a policy, but it guarded exactly one constant and would have sailed past `"Alex Dropbox"`. Narrower than it read.

What actually prevents a recurrence is already in the branch: the Browser header now uses `DROPBOX_CONNECTION_NAME` instead of keeping its own copy of the literal. That duplication was the real defect — it is why one personal string existed in two places and a rename could have missed the user-visible one. The doc comment on the constant records why the value matters.

If a guard is ever wanted for the general class, "no shipped user-visible string contains a maintainer's name" is a repo grep in CI, not a unit test. Not adding one on spec.

Tests still 412 passing, clippy and fmt clean.

PR #126, opened 2026-07-29
fix(piano-roll): land new notes in the cell the cursor is over

alexanderwanyoike:
Stack 1/3 — base `dev`. Independent of the other two.

## The bug

Clicking a cell in the piano roll sometimes created the note offset from where the cursor hit.

`SnapGrid::snap_beat` snaps to the **nearest** gridline, and note creation used it. So a click anywhere in the right half of a grid cell rounded *up* and created the note one cell further right than the pointer. At a 4-bar clip fitted to ~800px on a 1/16 grid the cells are ~12.5px, so anything within ~6px of a boundary landed in the wrong cell.

Nearest-line snapping is correct for *moving* and *resizing* an existing note. It is wrong for a creation gesture, which has to land where the user aimed. Every other DAW floors to the clicked cell.

## The fix

- Add `SnapGrid::snap_beat_floor` / `GridConfig::snap_beat_floor` for cell-containing snapping.
- Use it for the two creation paths only: draw-mode click and double-click add. Drag and resize keep round-to-nearest.

## Also

`hit_test_note` scanned notes forwards and returned the first match. Notes are painted in order, so the last one drawn sits on top — the hit test was returning the note *underneath* whenever two overlapped. Now scans back to front.

## Tests

Three new tests, all of which fail without this change:

- `creating_a_note_keeps_it_in_the_cell_the_pointer_is_over` — pins the round-vs-floor distinction directly (0.4 rounds to 0.5, floors to 0.25)
- `creation_snapping_is_free_when_snap_is_disabled`
- `overlapping_notes_hit_test_to_the_one_drawn_on_top`

`cargo build --workspace`, full test suite and `cargo clippy --all-targets` clean.

### DIFF
--- a/crates/vibez-ui/src/state/ui_types.rs
+++ b/crates/vibez-ui/src/state/ui_types.rs
@@ -510,6 +510,16 @@ impl SnapGrid {
         let size = self.beat_size();
         (beat / size).round() * size
     }
+
+    /// Snap a beat value down to the start of the grid cell containing it.
+    ///
+    /// Creation gestures must land in the cell the pointer is actually
+    /// over. `snap_beat` rounds to the nearest line, so a click in the
+    /// right half of a cell would create the note one cell too far right.
+    pub fn snap_beat_floor(self, beat: f64) -> f64 {
+        let size = self.beat_size();
+        (beat / size).floor() * size
+    }
 }
 
 impl std::fmt::Display for SnapGrid {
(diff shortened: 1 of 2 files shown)


alexanderwanyoike:
## Review

**Verdict: approve as-is.** Verified at the stack tip: `cargo build --workspace`, `cargo clippy --all-targets`, and the full test suite are green.

The domain reasoning is exactly right: creation gestures and move/resize gestures are different operations with different snapping semantics, and the code now names that distinction (`snapped_beat_floor` with a doc comment explaining the gesture it serves). A lazier fix would have changed `snap_beat` globally and broken drag feel.

- **Design:** the floor variant is added at both layers (`SnapGrid` and the `GridConfig` wrapper that owns the enabled check), mirroring the existing `snap_beat` pair exactly. Consistent extension, no modification of existing behavior, callers pick by intent.
- **Correctness:** floor is right for creation (a click exactly on a gridline starts there; anywhere inside a cell stays in it). The `max_start` guard still holds since floor never increases the beat. The reversed hit-test also means resize and drag now grab the visually top note, which is the desired side effect.
- **Tests:** the round-vs-floor test asserts both functions on the same input, which pins the distinction rather than just the new function. Nice.

One nit, not worth a respin: the two creation call sites still duplicate the `note_duration` / `max_start` guard block. Pre-existing, but if these paths get touched again, fold them into one helper.

PR #127, opened 2026-07-29
feat(timeline): box-select clips and notes by dragging

alexanderwanyoike:
Stack 2/3 — based on #126, **retarget to `dev` once that merges**.

Click-drag a box to select multiple clips (Arrange **and** Section construction) or multiple notes (piano roll).

## Approach

Dragging empty space (or a clip body) already set a horizontal time selection on one lane. Rather than add a second modal gesture, that same drag becomes a 2D rubber-band: it spans the lanes it is dragged across, selects every clip the box overlaps, and still carries the time selection when it has horizontal extent. A drag straight down has no range, so it leaves the existing one alone.

Sections come along for free — both surfaces render through the shared `TrackClipCanvas` and select into the shared `TimelineEditorState`, so one implementation covers both. They opt in via `with_marquee(...)`, keeping the already-28-argument `from_track` untouched.

## Row geometry, and a bug it fixed

Each lane is its own canvas, so the lane that starts the drag has to resolve the box on everyone's behalf. That needs to know where the other lanes actually are, added as `TrackRowSpan` (`widgets/timeline/marquee.rs`).

Cross-track clip drag now resolves through those same spans. **It was previously stepping by a hardcoded `70.0` row height**, so it targeted the wrong lane whenever a track above had its automation lanes expanded. `dev` had already made every automation row height an explicit constant, so the offsets are exact rather than approximated.

## Overlap vs containment

A clip or note counts as caught when its span *overlaps* the box — that is how a marquee reads on screen. The region ops (`op_delete_clips_in_region` and friends) deliberately keep demanding full containment, because deleting a clip you only clipped the edge of would destroy material outside the selection.

## Shift-drag

An additive drag unions onto the selection as it stood when the gesture *began*, captured on the first update and held in `ArrangementMarquee::base`. Without that, shrinking the box would never give anything back and the selection would only ever grow for the rest of the drag.

## Note on the first commit

`clips.rs` was already at 1,210 lines — over the repo's 1k ceiling — before this work. Its tests move to `clips_tests.rs` in a separate commit at the bottom, so the feature diff stays readable. No test bodies change. The file is now 965.

## Tests

- `marquee.rs` — 4 geometry tests, including uneven row heights (the automation-lane case that broke the old fixed-height math) and lane slicing
- arrangement domain — 5 tests: cross-lane coverage, edge-overlap, shrinking a shift-drag, time selection only on horizontal extent, end-of-drag
- piano roll domain — 3 tests: beat/pitch range, edge-overlap, additive vs replace

`cargo build --workspace`, all 24 suites and `cargo clippy --all-targets` clean.


alexanderwanyoike:
## Review

**Verdict: approve with three fix requests.** Verified at the stack tip: build, clippy, and all suites green. The fixes are all local to the new code, so the retarget plan in the PR body still holds.

### Strengths

- `widgets/timeline/marquee.rs` is a pure geometry module with its own unit tests, including the uneven-row-height case. Good SRP.
- The responsibility split follows Information Expert cleanly: the widget, which knows pixels, resolves the box into `track_ids` plus beats; the domain op receives musical/identity terms and never sees the layout. `op_marquee_select` sits in `ops.rs` with the other region ops and documents the overlap-vs-containment distinction against them.
- The shift-drag `base` capture is the correct algebra for additive marquees, and it is tested with the shrink case that breaks naive implementations.
- Reusing the existing `PendingSeek` to `RegionSelect` gesture instead of adding a modal tool keeps the interaction model small.
- Promoting the magic `70.0` to `TRACK_ROW_HEIGHT` and fixing the cross-track drag mis-targeting is exactly the "fix the neighbor while you are here" this codebase wants.

### Fix requests

1. **Selection and drawing disagree about the automation area** (`marquee.rs`: `resolve` vs `slice_for_lane`). `resolve()` includes a lane when the box intersects its *full* row, automation lanes included, but `slice_for_lane` deliberately clips drawing to the clip lane. Concrete failure: track 1 has automation expanded (row 70..196, clip lane 70..140); start a drag in track 2's lane and pull the top edge up to y=150, inside track 1's automation area. Track 1 lands in `track_ids`, its clips get selected, and no box is drawn on its lane, so clips select with no visual cause. One-line fix: filter on the lane rect (`row.top <= bottom_y && row.top + row.lane_height >= top_y`) so selection matches what `slice_for_lane` draws. The doc comment on `slice_for_lane` ("would imply the automation points are selected too") is the argument; `resolve` just does not follow it yet.

2. **The `row_spans` builder is duplicated verbatim** between `views_arrangement.rs` and `views_perform_sections.rs`, differing only in lane height and editor. This is exactly the geometry that must never drift between the two surfaces, which is the argument for it existing once. Extract a helper (natural home: next to `automation_rows_height`, or a builder in `marquee.rs` taking `(TrackId, extra_height)` pairs) parameterized by lane height.

3. **The widget carries the whole `ArrangementMarquee`, including `base`** (`with_marquee`, cloned per lane per frame in both views). The canvas only reads the four geometry fields to draw its slice; `base` is a `HashSet<ArrangementSelection>` that exists purely for the domain's additive-drag algebra, and after a select-all-then-shift-drag it can be large, cloned once per lane per frame at 60fps during the drag. This is an interface segregation problem before it is a perf one: the widget should depend on a small Copy rect (start/end beats, top/bottom y), not on the domain's gesture-state type. Splitting it also removes the odd situation where domain state permanently carries a field only widgets look at.

### Observations, no respin needed

- Arrange resolves the selection live during the drag; the piano roll resolves only on release (both draw the rectangle live). Worth aligning in a follow-up; live-select on the piano roll would just mean sending `SelectNotesInRegion` on move.
- `MARQUEE_MIN_PX = 4.0` is defined twice (`clips.rs`, `piano_roll/mod.rs`), and the two marquee draw blocks are near-identical with different fill alphas (0.10 vs 0.12). At minimum make the alpha intentional; a shared theme helper would cover both.
- Cross-track drag behavior change: the old code clamped to the first/last track when dragged beyond the column; `resolve_lane_at` now returns `None` outside the spans, so the clip stays in its lane. Probably the better behavior, but it is unstated in the PR body.
- `ArrangementMarquee.top_y/bottom_y` puts screen pixels into `TimelineEditorState`. Pragmatic tradeoff (lanes need to draw slices of one box) and the doc comment owns it, but it is the one place the pixel/beat boundary bends.

alexanderwanyoike:
## Re-review of 2fdde7b2

All three fix requests are resolved. **Approve.**

1. **Lane-rect filter:** `resolve()` now intersects against `row.top + row.lane_height` instead of the full row, matching `slice_for_lane`, and `a_box_over_only_automation_does_not_select_the_track` pins it. Selection and drawing can no longer disagree.
2. **Shared builder:** `build_row_spans` in `marquee.rs` single-sources the accumulation, callers keep only the surface-specific extra-height lookup, and the new accumulation test covers the uneven case. Exactly the right split.
3. **Rect split:** `ArrangementMarqueeRect` (Copy) for the widget, `base` stays in the domain's `ArrangementMarquee`. The per-lane `HashSet` clone is gone and the widget no longer depends on gesture state.

Verified at the rebased stack tip: build, clippy, and all suites green.

One cosmetic leftover, not worth a push on its own: `views_arrangement.rs` imports `build_row_spans` while `views_perform_sections.rs` calls it fully qualified. Align whenever you next touch either file.

PR #128, opened 2026-07-29
feat(timeline): select all with Command+A

alexanderwanyoike:
Stack 3/3 — based on #127, **retarget to `dev` once the stack merges**.

Command+A selects every clip on the active timeline (Arrange and Section construction), or every note when the piano roll is the visible editor.

## Why it is routed centrally

Command+A had no global binding at all. The piano roll canvas bound it locally — which is the exact race the existing `DeleteKeyPressed` comments warn about:

> every canvas receives keyboard events, so two canvases binding the same key race each other

So this follows the precedent Delete already set: one `Message::SelectAllPressed` from `global_key_handler`, context-resolved in `update`. The local canvas binding is removed.

Because resolution goes through `active_timeline_editor()`, Arrange and Sections are both covered by the same handler.

## The one difference from Delete

`on_delete_key_pressed` decides between notes and clips by asking *what is already selected*. Select-all cannot do that — creating a selection is the whole point, so there is never one to inspect. Instead `open_piano_roll_clip()` asks whether the piano roll is actually on screen, mirroring the condition `views_detail` uses to decide whether to draw it (Clip tab + MIDI track + selected note clip on the selected track).

Text fields keep Command+A for their own contents, same guard as Delete.

## Tests

- `command_a_routes_centrally_instead_of_binding_a_canvas` — asserts the shortcut is surface-agnostic, and that bare `a` stays a Perform pad key
- `select_all_clips_takes_every_clip_on_every_track`
- `select_all_clips_on_an_empty_timeline_selects_nothing` — replaces rather than extends a stale selection, and does not pull focus to a clip tab with nothing in it

`cargo build --workspace`, all 24 suites and `cargo clippy --all-targets` clean.

### DIFF
--- a/crates/vibez-ui/src/app/keyboard.rs
+++ b/crates/vibez-ui/src/app/keyboard.rs
@@ -438,6 +438,7 @@ pub(crate) fn global_key_handler(
             Some(Message::Arrangement(ArrangementMsg::MoveSelectedTrackDown))
         }
         iced::keyboard::Key::Character(ref c) => match c.as_str() {
+            "a" | "A" => Some(Message::SelectAllPressed),
             "c" | "C" => Some(Message::Arrangement(ArrangementMsg::CopySelectedClips)),
             "x" | "X" => Some(Message::Arrangement(ArrangementMsg::CutSelectedClips)),
             "v" | "V" => Some(Message::Arrangement(ArrangementMsg::PasteClips)),
@@ -490,6 +491,25 @@ mod tests {
         {
             iced::keyboard::Modifiers::CTRL
         }
+    }
+
+    #[test]
+    fn command_a_routes_centrally_instead_of_binding_a_canvas() {
+        use iced::keyboard::{Key, Modifiers};
+
+        // Context resolution happens in update(); the shortcut itself is
+        // surface-agnostic so the piano roll and the timeline canvases
+        // cannot race each other for it.
+        assert!(matches!(
+            global_key_handler(Key::Character("a".into()), command_modifier()),
+            Some(Message::SelectAllPressed)
+        ));
+        assert!(matches!(
+            global_key_handler(Key::Character("A".into()), command_modifier()),
+            Some(Message::SelectAllPressed)
+        ));
+        // Bare "a" is a Perform pad key, not select-all.
+        assert!(global_key_handler(Key::Character("a".into()), Modifiers::empty()).is_none());
     }
 
     #[test]
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -604,6 +604,7 @@ impl App {
 
             // -- Multi-track messages --
             Message::DeleteKeyPressed => return self.on_delete_key_pressed(),
+            Message::SelectAllPressed => return self.on_select_all_pressed(),
             Message::AddClipToTrack(track_id) => {
                 return self.handle_add_clip_to_track(track_id);
             }
--- a/crates/vibez-ui/src/app/update_media.rs
+++ b/crates/vibez-ui/src/app/update_media.rs
@@ -19,13 +19,16 @@ use crate::message::{BrowserImportTarget, Message};
 use super::*;
 
 impl App {
+    fn text_field_editing(&self) -> bool {
+        self.state.view.editing_track_name.is_some()
+            || self.state.view.editing_clip_name.is_some()
+            || self.state.perform.editing_section_name.is_some()
+    }
+
     pub(super) fn on_delete_key_pressed(&mut self) -> Task<Message> {
         // Never delete anything while a text field is being
         // edited; backspace belongs to the text there.
-        if self.state.view.editing_track_name.is_some()
-            || self.state.view.editing_clip_name.is_some()
-            || self.state.perform.editing_section_name.is_some()
-        {
+        if self.text_field_editing() {
             return Task::none();
         }
         // Priority 1: a selected automation point.
@@ -68,10 +71,7 @@ impl App {
     /// piano roll is on screen instead.
     pub(super) fn on_select_all_pressed(&mut self) -> Task<Message> {
         // A text field owns Command+A for its own contents.
-        if self.state.view.editing_track_name.is_some()
-            || self.state.view.editing_clip_name.is_some()
-            || self.state.perform.editing_section_name.is_some()
-        {
+        if self.text_field_editing() {
             return Task::none();
         }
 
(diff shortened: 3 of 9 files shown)


alexanderwanyoike:
## Review

**Verdict: approve with two fix requests.** Verified at the stack tip: build, clippy, and all suites green.

The centralization is correct and well-argued; the removed canvas binding with the explanatory comment left in its place is a nice touch, and `SelectAllClips` living in the domain with the empty-timeline/stale-selection test is right.

### Fix requests

1. **The text-field guard is now copy-pasted** between `on_delete_key_pressed` and `on_select_all_pressed` in `update_media.rs`, three conditions each, and the next global shortcut will make a third copy. Extract `fn text_field_editing(&self) -> bool` on `App` and use it in both.

2. **`open_piano_roll_clip` is an inexact mirror of `views_detail`, and the drift is already observable in this very stack.** `views_detail` shows the piano roll when the track has a `NoteClip` in `selected_clips` *or* a matching `selected_note_clip`; `open_piano_roll_clip` checks only `selected_note_clip`. #127's marquee populates `selected_clips` with `NoteClip` entries but never sets `selected_note_clip`, so: marquee-select a note clip on the selected MIDI track, the detail panel enters its piano roll branch, press Command+A, and you get all *clips* selected instead of all notes. The behavior is defensible, but two hand-maintained copies of "is the piano roll the visible editor" is the real problem. Extract one predicate on `App` and use it from both `views_detail` and the shortcut resolution; then the answer to "which surface owns Command+A" can never disagree with what is on screen.

### Inherited gap, suggest filing an issue rather than blocking

The guard covers the three name-editing states, but the app has several other live text inputs: browser search, automation parameter search, context-menu search, BPM and swing fields. Command+A while typing in the browser search will also select every clip and yank the detail panel to the clip tab via `focus_clip_tab`. Delete has the same hole (worse: it deletes clips while you edit a search string), so this PR is consistent with precedent, but Command+A inside a text field is a much more common gesture than Delete, which makes the gap more visible now. The durable fix is a single "a text field owns the keyboard" predicate fed by actual focus tracking, applied to both shortcuts.

alexanderwanyoike:
## Re-review of 86907f33

Both fix requests are resolved. **Approve.**

1. **`text_field_editing()`** extracted and used by both shortcuts.
2. **Single resolver:** `open_piano_roll_clip` now lives in `views_detail.rs` and drives rendering (both the tab branch and `view_piano_roll_panel`'s clip data) and shortcut routing, with the precedence test. This went further than asked in a good way: a marquee-selected note clip now actually renders in the piano roll instead of an empty shell, so the panel and Command+A can no longer disagree.

Verified at the stack tip: build, clippy, all suites green.

Two behavioral notes to be aware of, neither blocking:

- **Arbitrary pick under multi-selection:** `selected_note_clip_for_track` falls back to `find_map` over the `HashSet`, so when several note clips on the track are selected and none is explicitly open (wide marquee, or select-all), which clip the piano roll shows is unspecified. Stable within a gesture, but a deterministic tie-break (earliest `position_beats` via a timeline lookup) would make the behavior explainable.
- **Command+A now escalates:** the first press selects all clips and focuses the Clip tab; if the selected track is MIDI, the resolver then finds one of its now-selected note clips, so a second press selects all notes inside that clip. Ableton-ish and arguably nice, but it composes with the arbitrary-pick note above, so worth a conscious decision.

Cosmetic: the blank line between `on_select_all_pressed` and `on_load_sampler_sample` got dropped in `update_media.rs`.

alexanderwanyoike:
## Follow-up: Command+A must not escalate

Resolving the "Command+A now escalates" note from the re-review: after discussion, escalation is wrong and should be removed. Ableton is the reference behavior here.

### The rule

Ableton routes Cmd+A by *focus*, the editor the user last clicked in, not by what is visible on screen. Cmd+A in the Arrangement selects all arrangement material and pressing it again does nothing new; the selection it just created never moves focus into the MIDI editor. Select-all must never change its own target.

### Why the current code escalates

`open_piano_roll_clip` treats "a note clip is in `selected_clips`" as evidence the piano roll is the active editor. After the first Cmd+A that is true for every note clip on the selected track, so the second press re-routes to notes, and via the `find_map` fallback it picks an arbitrary clip to boot. Selection state is being used as a proxy for focus, and select-all pollutes that proxy.

### The fix

Route the shortcut on the *explicit* open only. Split the shared resolver into two thin layers over the same tab/track/MIDI conditions:

- `visible_piano_roll_clip()`: current behavior including the `selected_clips` fallback; used by `views_detail` to decide what to draw.
- `focused_piano_roll_clip()`: same checks but resolves through `selected_note_clip` only; used by `on_select_all_pressed`.

This yields exactly Ableton's semantics:

- Cmd+A twice is idempotent (second press re-runs `SelectAllClips`, same result).
- Marquee-selecting a note clip still renders its notes, but Cmd+A selects clips, matching Ableton where marquee focus stays in the arrangement.
- Clicking a note clip open (`selected_note_clip` set) then Cmd+A selects its notes.
- The arbitrary `find_map` pick no longer decides which clip's notes get selected, because routing never uses the fallback.

Do **not** implement this as "detect and ignore a repeated Cmd+A"; that special-cases the symptom and would still escalate on the third press after any selection change.

### Note on the earlier unification

This deliberately walks the resolver unification back half a step, and that is fine: rendering and routing answer two genuinely different questions ("what can I show?" vs "where is the user working?"). Keep them as two named views over one shared set of conditions, not two hand-maintained copies.

### Tests

Adjust `select_all` coverage to pin no-escalation: with all clips selected and no explicit `selected_note_clip`, a second Cmd+A selects clips again, not notes. Keep a positive case: with `selected_note_clip` set and the piano roll visible, Cmd+A selects all notes in that clip.

alexanderwanyoike:
## Re-review of 835f2abd

**Approve.** This implements the no-escalation spec exactly, and the shape is right: `piano_roll_track()` holds the shared structural gate once, and `visible_piano_roll_clip()` / `focused_piano_roll_clip()` are two named views over it rather than two hand-maintained copies. The doc comments name the distinction ("may supply the clip without transferring keyboard focus") so the next reader cannot re-merge them by accident.

Verified against every requirement from the spec comment, at the tip (build, clippy, fmt, all suites green, 439 vibez-ui tests):

- Cmd+A twice is idempotent: the focused resolver ignores `selected_clips`, so the second press re-runs `SelectAllClips`.
- Marquee-selected note clip renders but does not capture Cmd+A.
- Explicitly opened clip (`selected_note_clip`) gets all-notes.
- No repeat-detection hack; the fix is in routing, not in the gesture.
- The arbitrary `find_map` pick is now rendering-only.
- `arrangement_selection_does_not_focus_the_piano_roll_for_select_all` pins the exact post-first-press state, with the positive explicit-open case alongside.

One optional nit: `visible_note_clip_for_track` re-inlines the explicit branch that `focused_note_clip_for_track` owns. Writing it as `focused_note_clip_for_track(editor, track_id).or_else(|| fallback)` would encode "focused implies visible" structurally instead of by parallel code. Two-line change, fold it in whenever.

One known limitation to leave documented, not fix: `selected_note_clip` means "explicitly opened at some point", not "last interacted". Open a note clip, then marquee on that same track, and Cmd+A still targets the notes because nothing in the arrangement gesture clears the open clip. That is defensible while the roll is visibly open, and the case is narrow (same-track anchor only, since the marquee moves `selected_track`). If more focus-routed shortcuts accumulate, the durable answer is a real focus field in state that every surface interaction updates; do not grow more selection-state proxies.

With this, the whole stack is approve from my side: #126, #127, and #128 all clear.

PR #134, opened 2026-07-30
feat(ui): add an About dialog

alexanderwanyoike:
## Why

vibez ships nightly-ish builds from several places (packages, tarballs, the releases page) and there was no way to answer "which build is this?" from inside the app. `--version` on the CLI exists, but a producer who launched from a desktop icon never sees it. Bug reports carry no provenance.

## What lands

`File > About vibez` opens a modal with:

- app name and the version line, e.g. `0.1.2 (1ed3e55)`
- the license, `GPL-3.0-or-later`, matching LICENSE and the workspace manifest
- Repository, Website and Releases links, each rendered as label plus its URL so the destination is visible before the click, since a browser handoff cannot be previewed or undone

## The commit hash

A new `crates/vibez-ui/build.rs` bakes `git rev-parse --short=7 HEAD` in as `VIBEZ_GIT_HASH`. It lives in vibez-ui rather than the root package because a build script's `rustc-env` only reaches the crate it belongs to, and the dialog is drawn there.

Every failure path degrades to `"unknown"` instead of aborting: no `.git` (release tarball, distro package, crates.io), no `git` binary, or a failed invocation. A missing commit is a cosmetic loss in one dialog, never a reason for vibez to fail to compile. `about::version_line` then drops an unknown commit entirely rather than printing `0.1.2 (unknown)`, which reads to a user as a defect where a bare `0.1.2` correctly reads as a release build.

Rerun tracking watches both `HEAD` and the current branch ref, resolved through `git rev-parse --git-path` so it is correct inside a linked worktree, so the baked hash follows a checkout *and* a fresh commit on the same branch. Paths that do not exist are skipped, because naming a missing file makes cargo rebuild vibez-ui on every invocation.

Verified by compiling the script standalone: inside the repo it emits `VIBEZ_GIT_HASH=1ed3e55` plus two rerun paths, and in a directory outside any repository it emits `VIBEZ_GIT_HASH=unknown` and no rerun paths.

## No new dependency

vibez-dropbox already owns `open = "5"` behind its `BrowserOpener` trait for the OAuth flow, and vibez-ui already constructs `SystemBrowserOpener` in `async_helpers.rs`. Link presses reuse it. `BrowserOpener::open` waits for the platform launcher to return, so it runs on `spawn_blocking`: on a cold browser start that stall would otherwise freeze the UI. A failure surfaces in the status bar rather than being swallowed.

## Dismissal

The dialog is registered as `MenuOverlay::About` rather than following the Settings modal. Settings can only be closed by its X or a backdrop press; Escape does not reach it. Going through `menu_lifecycle` means the close button, the backdrop and Escape all retire the About dialog through the one path that module exists to own, at no extra cost. It sits above the File menu in both `visible()` and the shell's overlay chain, matching the existing ordering rule.

Placement is the File menu, next to Settings, rather than a new Help menu: File is where this app already keeps application-level entries, and a menu holding a single item is not worth the header space.

## Tests

- `crates/vibez-ui/src/about.rs`: `the_version_line_names_the_commit_a_build_came_from`, `a_commit_no_build_script_could_resolve_is_dropped_not_shown`, `the_version_line_survives_the_whitespace_git_output_carries`, `the_running_build_reports_its_own_cargo_version`.
- `crates/vibez-ui/src/app/menu_lifecycle.rs`, in the existing style: `the_about_dialog_outranks_the_menu_it_was_opened_from`, `the_about_item_opens_the_dialog_and_closes_the_file_menu`, `escape_closes_the_about_dialog`.

The view itself is pure layout and is untested, matching the other `views_*` modules.

## Verification

`cargo build --workspace`, `cargo clippy --workspace --all-targets`, `cargo test --workspace` (450 passing in vibez-ui) and `cargo fmt --check` all pass. Clippy reports only the two pre-existing raw-pointer-cast warnings in vibez-plugin-host, untouched here.

Every touched file stays under the 1,000-line ceiling. The dialog is a new `views_about.rs` (128) because `views_settings.rs` is already at exactly 1,000. Worth flagging: `state/mod.rs` is now 996 after one added field, so the next change there should split it.

### DIFF
--- a/crates/vibez-ui/build.rs
+++ b/crates/vibez-ui/build.rs
@@ -1,0 +1,57 @@
+//! Bakes the short commit of the build tree into the binary so the About
+//! dialog can name the exact source a build came from.
+//!
+//! Release tarballs, distro packages and crates.io builds ship without a
+//! `.git` directory, and some build machines have no `git` at all. Every
+//! failure path here therefore degrades to `UNKNOWN_COMMIT` instead of
+//! aborting the build: a missing commit is a cosmetic loss in one dialog,
+//! never a reason for vibez to fail to compile.
+
+use std::path::PathBuf;
+use std::process::Command;
+
+/// Must stay in step with `crate::about::UNKNOWN_COMMIT`; a build script
+/// cannot import from the crate it builds.
+const UNKNOWN_COMMIT: &str = "unknown";
+
+fn main() {
+    let commit = short_commit().unwrap_or_else(|| UNKNOWN_COMMIT.to_string());
+    println!("cargo:rustc-env=VIBEZ_GIT_HASH={commit}");
+
+    // Watch HEAD and the branch ref so the baked hash follows both a
+    // checkout and a fresh commit on the current branch. Paths that do not
+    // exist are skipped: naming a missing file makes cargo rebuild this
+    // crate on every single invocation.
+    for path in head_watch_paths() {
+        println!("cargo:rerun-if-changed={}", path.display());
+    }
+}
+
+fn short_commit() -> Option<String> {
+    let hash = git(&["rev-parse", "--short=7", "HEAD"])?;
+    (!hash.is_empty()).then_some(hash)
+}
+
+/// Files whose content changes whenever the checked-out commit changes.
+fn head_watch_paths() -> Vec<PathBuf> {
+    let mut refs = vec!["HEAD".to_string()];
+    if let Some(branch) = git(&["symbolic-ref", "--quiet", "HEAD"]) {
+        refs.push(branch);
+    }
+
+    refs.iter()
+        .filter_map(|reference| git(&["rev-parse", "--git-path", reference]))
+        .map(PathBuf::from)
+        .filter(|path| path.exists())
+        .collect()
+}
+
+/// Trimmed stdout of a successful `git` run, or `None` when git is absent,
+/// this is not a repository, or the command failed.
+fn git(args: &[&str]) -> Option<String> {
+    let output = Command::new("git").args(args).output().ok()?;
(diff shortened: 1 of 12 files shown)


alexanderwanyoike:
## Review

**Approve.** Independently verified: 450 vibez-ui tests green on the branch; diff read in full.

What is done well:

- **No new dependency**: reusing `vibez_dropbox::SystemBrowserOpener` on `spawn_blocking`, with the stall rationale documented, is exactly the reuse-first answer.
- **`build.rs` is properly paranoid**: every failure path degrades to `unknown`, the About line then drops the commit instead of printing "0.1.2 (unknown)", and the rerun tracking resolves refs through `git rev-parse --git-path` so worktrees stamp correctly, skipping missing paths to avoid perpetual rebuilds. The "must stay in step with `about::UNKNOWN_COMMIT`" comment marks the one cross-file constant honestly.
- **Dismissal through `MenuOverlay::About`** rather than a copied modal means Escape, backdrop, and the close button all retire it through the single `menu_lifecycle` path, with the overlay-precedence test (`the_about_dialog_outranks_the_menu_it_was_opened_from`) pinning the stacking.
- `views_about.rs` as a new module because `views_settings.rs` sits at the 1,000-line ceiling: correct call.

One shared quirk, not a blocker: the backdrop `mouse_area` wraps the dialog too, so a click on the card's non-interactive interior dismisses it. This faithfully copies the settings modal's existing pattern, so consistency wins here, but both modals share the flaw. Worth one tiny follow-up that swallows presses on the card for both.

Noted for the merge queue: this PR and #135 both touch `views_shell.rs` and `message.rs`; whichever merges second gets a trivial conflict.

PR #135, opened 2026-07-30
feat(app): title the window after the open project and confirm on close

alexanderwanyoike:
## Why

Two gaps that show up the moment you have more than one project on disk.

The window title was the literal string `"vibez"`. With several vibez windows open, or just one open next to everything else in a taskbar, nothing said which project you were looking at. The dirty flag already existed and was already surfaced, but only inside the File menu as a `Save*` label, which means you had to open a menu to learn you had unsaved work.

Closing the window exited immediately. `route_new_project` discards without asking too, but that at least keeps you in the app; the close button lost the session outright.

## Title

`iced::application(...)` now takes `App::title` instead of `"vibez"`. iced re-evaluates it after every update, so the title tracks state without anything having to push it.

The format is `vibez - {name}`, `" *"` appended while dirty, and `Untitled` for a project that has never been saved.

The name is the path's file stem, not the full path, so the project name stays the part of the title that taskbars truncate last. That derivation already existed inline in `project_from_state`, which is what writes `Project.name` into the saved file. Rather than write a second copy, both now call `window_policy::project_display_name`, so the title bar and the serialized project cannot disagree about what an unsaved project is called. This is the only change outside the feature and it makes `project_io.rs` shorter.

No new dirty tracking was introduced. `state.project.dirty` and `state.project.current_path` are read as they are.

## Confirm on close

`.exit_on_close_request(false)` on the builder, plus a `window::Event::CloseRequested` arm in the existing `listen_with` subscription, so every close request lands in `update` as `Message::WindowCloseRequested`.

A clean project closes with no dialog. A dirty one raises the overlay. The decision is a one-argument function over the dirty flag rather than a check spread through the handler, which also means a repeated close request while the dialog is already up re-affirms `Confirm` instead of racing past the prompt.

The Save path deliberately does not reimplement saving. It sets `exit_after_save` and dispatches the existing `Message::SaveProject`, which already falls through to save-as when there is no path yet. That intent has to be a state field rather than a local, because the save is asynchronous and may open a native file dialog first. Three ways out of it, all of which leave you in the app with your edits:

- the save-as dialog is cancelled (`route_project_save_path_selected(None)`),
- the write fails, in which case the error stays on screen,
- Cancel is pressed before any of it starts.

The overlay is inserted at the **top** of the overlay chain in `views_shell.rs`: whatever menu happened to be open, "are you about to lose work" has to be the visible question.

## Styling

Matches `view_track_deletion_overlay` exactly: 440px card, `PERFORM_DISPLAY` title at 18, body copy at 11 in `text_dim`, buttons right-aligned behind a `horizontal_space` with `padding([7, 14])`, backdrop at 62% black. Discard carries the `th::danger()` styling that Delete carries elsewhere, since discarding is the destructive action here, not closing. Save carries the accent.

It lives in its own `views_close_confirm.rs` rather than in `views_overlays.rs`, which was at 898 lines and had no room left under the 1,000-line ceiling.

## Tests

`crates/vibez-ui/src/app/window_policy.rs`, testing the pure rules directly since `App` owns an audio engine and cannot be constructed in a test:

- `a_saved_project_titles_with_its_file_stem_not_its_path`
- `a_never_saved_project_titles_as_untitled`
- `unsaved_changes_append_the_dirty_marker_for_named_and_untitled_projects`
- `the_dirty_marker_disappears_once_the_project_is_saved`
- `a_path_without_an_extension_still_titles_with_its_name`
- `a_clean_project_closes_without_a_prompt`
- `unsaved_changes_turn_a_close_request_into_a_confirmation`

## Verification

`cargo build --workspace`, `cargo clippy --workspace --all-targets`, `cargo test --workspace`, and `cargo fmt --check` all pass. Clippy under CI's `-D warnings` is clean; the only warnings anywhere are the two pre-existing raw-pointer casts in `vibez-plugin-host`, untouched here.

Every file touched stays under the 1,000-line ceiling. `project_io.rs` was already over at 1240 and this change takes it down to 1235; splitting it is out of scope for this one.

### DIFF
--- a/crates/vibez-ui/src/app/mod.rs
+++ b/crates/vibez-ui/src/app/mod.rs
@@ -102,10 +102,13 @@ pub fn run() -> iced::Result {
         64,
     )
     .ok();
-    iced::application("vibez", App::update, App::view)
+    iced::application(App::title, App::update, App::view)
         .theme(App::theme)
         .antialiasing(true)
         .subscription(App::subscription)
+        // Unsaved edits get a confirmation instead of a silent exit, so the
+        // close request has to reach `update` rather than the window itself.
+        .exit_on_close_request(false)
         .window({
             #[allow(unused_mut)]
             let mut settings = iced::window::Settings {
@@ -169,6 +172,7 @@ mod views_browser_places;
 mod views_browser_remote;
 mod views_browser_style;
 mod views_clip_swing;
+mod views_close_confirm;
 mod views_detail;
 mod views_devices;
 mod views_mixer;
@@ -183,6 +187,7 @@ mod views_settings;
 mod views_settings_perform;
 mod views_shell;
 mod views_transport;
+mod window_policy;
 
 #[cfg(test)]
 mod project_format_v1_tests;
@@ -545,6 +550,15 @@ impl App {
         }
     }
 
+    /// Re-evaluated by iced after every update, so the title tracks the open
+    /// project and its dirty flag without anything having to push it.
+    fn title(&self) -> String {
+        window_policy::window_title(
+            self.state.project.current_path.as_deref(),
+            self.state.project.dirty,
+        )
+    }
+
     fn theme(&self) -> Theme {
         th::vibez_theme()
     }
@@ -566,6 +580,9 @@ impl App {
                 )),
                 iced::Event::Window(iced::window::Event::Unfocused) => {
                     Some(Message::Perform(PerformMsg::WindowUnfocused))
+                }
+                iced::Event::Window(iced::window::Event::CloseRequested) => {
+                    Some(Message::WindowCloseRequested)
                 }
                 _ => None,
             }),
--- a/crates/vibez-ui/src/app/project_io.rs
+++ b/crates/vibez-ui/src/app/project_io.rs
@@ -371,14 +371,9 @@ impl App {
             .collect();
 
         Project {
-            name: self
-                .state
-                .project
-                .current_path
-                .as_ref()
-                .and_then(|path| path.file_stem())
-                .map(|name| name.to_string_lossy().to_string())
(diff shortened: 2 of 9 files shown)


alexanderwanyoike:
## Review

**Approve.** Independently verified: 450 vibez-ui tests green on the branch; diff read in full.

What is done well:

- **`window_policy.rs`** as a pure module with the stated reason (App owns an audio engine and cannot be built in tests) is the right shape, and the tests read as specification.
- **The duplication fix nobody asked for**: `project_from_state` had the same file-stem/"Untitled" chain inline; both now share `project_display_name`, so the title bar and the serialized `Project.name` cannot disagree. This is the kind of neighboring cleanup this repo wants.
- **`exit_after_save` lifecycle is airtight**: set only by the dialog's Save, consumed on successful save, cleared on save error (with the edits intact and the error visible), and cleared when the save-as dialog is cancelled. Every abort path was considered.
- Close-confirm outranking every other overlay in `views_shell` with the justification in place, and repeated close requests re-affirming `Confirm` so hammering the close button cannot race past the prompt.

Two follow-ups, neither blocking:

1. **Escape should mean Cancel** while the confirm dialog is up. The dialog is not a `MenuOverlay`, so Escape currently does nothing there (or worse, acts on something underneath). Either register it in `menu_lifecycle` the way #134's About did, or special-case Escape while `close_confirm_open`.
2. **New and Open are still unguarded**: File > New or opening another project while dirty silently loses edits, the same hole this PR just closed for window close. `close_request_decision` is already the reusable policy; a follow-up PR should route those two flows through the same confirm (with "Save / Discard / Cancel" then continuing the original action instead of quitting).

Noted for the merge queue: this PR and #134 both touch `views_shell.rs` and `message.rs`; whichever merges second gets a trivial conflict.

PR #136, opened 2026-07-30
feat(settings): configurable interface scale

alexanderwanyoike:
## Why

vibez draws at whatever size the display's own scaling gives it, and nothing in the app changes that. On a 4K laptop panel the mixer strips and clip names are small enough to squint at; from across the room, performing off a screen a metre away, they are unreadable. Timeline zoom is not the answer to either: it changes how much musical time a lane shows, not how big the interface is drawn, and resizing the window just reveals more of the same small controls.

## The change

Appearance gains an **Interface scale** slider, 75% to 150% in 5% steps, default 100%. It is wired to iced's `scale_factor` hook on the application builder, which multiplies the logical coordinate space the entire interface lays out in, so every panel, control and font moves together in one step.

Naming is deliberate: "Interface scale" in the settings, `interface_scale` in the code, kept lexically distinct from timeline zoom so the two never read as the same knob. No zoom code is touched.

The builder edit is a single line, `.scale_factor(App::scale_factor)`, because #135 is concurrently changing the title function and `exit_on_close_request` on the same call. The two should merge without a conflict resolution.

## Persistence and the clamp

The value rides the existing `UiSettings` mechanism: a `#[serde(default = "...")]` field in `ui.json`, hydrated in `App::new`, rebuilt in `persist_ui_settings`, written to disk immediately on slider input the way `SetWarpConfidenceThreshold` does.

Clamping happens twice, on load and on input, and that is not belt-and-braces. `ui.json` is plain JSON with no schema version and people do hand-edit it. A scale of `0.05` or `40.0` from a typo would open the window at a size where the settings modal itself cannot be read or reached, so there would be no way to undo the mistake from inside the app. Clamping on hydrate makes that unreachable, in the same spirit as the existing `sample_browser_width` clamp.

`NaN` is handled separately rather than passed to `f32::clamp`, which returns `NaN` unchanged. A `NaN` scale factor collapses layout entirely, so an uninterpretable value falls back to unscaled.

## Canvas hit-testing

Worth stating explicitly, since this change moves the coordinate space every canvas in the app draws into. iced applies the scale factor above the widget layer: widgets lay out and receive events in logical units, and the cursor position a `canvas::Program` sees is already descaled. So the timeline, piano roll, meters and knobs need no per-widget adjustment, and pixel constants like `MEDIA_DRAG_THRESHOLD_PX` stay correct in logical terms.

One consequence is worth a note rather than a pre-emptive fix: a drag threshold expressed in logical pixels now corresponds to more physical movement at 150% and less at 75%. That is arguably the right behaviour, since the widgets it guards also grew, and it is what every other logical-unit constant in the app does. Flagging it so that if drag feel does turn out to be off at the extremes, this is the place to look, rather than tuning thresholds blind now.

## File split

`views_settings.rs` sat at exactly the 1,000-line ceiling, so the first commit moves the Appearance tab whole into `views_settings_appearance.rs`, the way `clips_tests.rs` was split from `clips.rs`. The tab is self-contained: it reads only theme state and emits only theme messages. That commit is otherwise mechanical, apart from replacing an em dash in the "no themes found" hint with a colon, since em dashes are banned in this codebase. The feature diff then sits on top and stays readable.

`views_settings.rs` is now 807 lines and `views_settings_appearance.rs` is 245. `state/mod.rs`, which is also near the ceiling, is untouched: the scale lives on the `App` shell rather than in `AppState`, because it describes the window the project is viewed through, not the project or the view layout, and nothing about it is undoable.

## Tests

Four behaviour-named tests in `ui_settings.rs`, none of which touch the real filesystem:

- `settings_written_before_interface_scale_render_unscaled` - legacy `{}` gets 100%.
- `interface_scale_roundtrips` - serde round-trip at 1.25.
- `a_hand_edited_interface_scale_is_pulled_back_into_the_usable_range` - 0.05, 40.0 and -2.0 land on the bounds.
- `an_uninterpretable_interface_scale_falls_back_to_unscaled` - NaN goes to 100%, infinities to the bounds.

## Verification

`cargo build --workspace`, `cargo clippy --workspace --all-targets`, `cargo test --workspace` and `cargo fmt --check` all pass. Clippy reports only the two pre-existing raw-pointer-cast warnings in `vibez-plugin-host`, untouched here. 447 tests pass in `vibez-ui`, including the four new ones.

Not verified by running the app: this machine's disk hit 100% during the work and there was no room to launch a GPU-backed window. The wiring is a single documented iced hook, and everything it depends on is covered above, but a reviewer pulling the branch should drag the slider to both ends once.


alexanderwanyoike:
## Review

**Approve.** Independently verified: 447 vibez-ui tests green on the branch, including the four new interface-scale tests; diff read in full.

What is done well:

- **The split-first discipline**: the Appearance tab moved out of `views_settings.rs` (1,000 -> 807 lines) as its own mechanical commit before the feature, exactly how the file-size ceiling is supposed to be maintained. The one non-mechanical edit in the move (an em dash swapped for a colon) is even a repo-rule fix.
- **Placement argument is right**: `interface_scale` on the `App` shell, not `AppState`, with the reasoning in the doc comment. It describes the window, not the project or the view layout, and it is not undoable.
- **The clamp is genuinely defensive**: explicit NaN handling because `f32::clamp` passes NaN through and a NaN scale factor bricks the window with no in-app recovery; infinities pin to the range ends; and both bounds carry their constraint in the doc comment (min: hit targets stop being aimable; max: the arrangement stops fitting the 900x600 minimum window). Clamped on hydrate and on input.
- **Merge-friendliness honored**: the builder change is the single `.scale_factor(App::scale_factor)` line, so the conflict with #135's builder edits is trivial.
- Tests match the house idiom exactly: legacy `{}` default, round-trip, hand-edited out-of-range, uninterpretable values.

One observation, not a fix request: `SetInterfaceScale` fires per slider tick and each tick calls `persist_ui_settings()`, a full non-atomic `ui.json` rewrite. The warp-confidence slider already behaves this way, so this is house pattern, but it is now two sliders amplifying the same known weakness (non-atomic settings save, noted previously). If settings saves ever grow a debounce or an atomic temp-and-rename like `RemoteCatalogStore` has, both sliders benefit.

Merge-queue note: #135 and #136 both edit the `iced::application` builder chain; whichever lands second has a one-line conflict.

PR #137, opened 2026-07-30
feat(updates): notify passively when a newer release exists

alexanderwanyoike:
## What

Vibez now tells you when a newer version has been released. On startup it asks GitHub for the latest release tag and, if it is newer than the running build, raises a dismissible chip in the status bar with a link to the releases page.

Nothing is downloaded and nothing is installed. The only action offered is opening a browser.

## Why this shape

A DAW is routinely run on machines with no network, so a release check that could not complete is not a problem the user needs to hear about. Every failure path is silent: DNS, TLS, timeouts, rate limits and malformed payloads all collapse to "nothing to report". An offline machine behaves exactly like one that opted out.

The notice is a status-bar chip rather than a dialog, so learning about a new version never steals focus from the arrangement or interrupts playback.

## Decisions

- **HTTP client**: reuses the `reqwest` + `rustls-tls` client the Dropbox integration already links, so the release check adds no new transitive stack.
- **Browser opener**: reuses `vibez-dropbox`'s existing `SystemBrowserOpener` on `spawn_blocking`, so there is no second dependency on `open` and the UI thread never waits on a browser launch.
- **Threading**: the fetch runs through `Task::perform`, so neither the UI nor the audio thread ever blocks on it. A 10s request timeout keeps a half-open connection from parking a runtime worker for the session.
- **Version comparison**: numeric per component, tolerating a leading `v`. Anything else yields no notice, pre-release suffixes included: an unparseable tag is indistinguishable from a tagging scheme this build predates, and a passive notifier should not nudge stable users onto release candidates.
- **Throttle**: at most one check per 24 hours. The timestamp is recorded for failed attempts too, so a laptop that is offline every morning backs off instead of retrying on every launch.
- **Persistence**: `check_for_updates` (default on) and `last_update_check_unix` join `UiSettings` in `ui.json`, both with per-field serde defaults so existing settings files load unchanged.
- **Dismissal**: session-scoped. The next launch re-raises it, which keeps a genuinely stale install from going unnoticed forever, while a release newer than the dismissed one re-raises it immediately.

## Settings

A new Settings > Updates tab carries "Check for updates on startup", default on. When off, no request is made at all. It also reports the running version and whatever the last check found.

## File split

`views_settings.rs` sat at the repository's 1,000-line ceiling. The first commit moves `view_settings_dropbox_tab` verbatim into `views_settings_dropbox.rs`, mirroring the existing `views_settings_perform` split, to make room. No behaviour change.

## Tests

18 new tests, none touching the network or the filesystem:

- Version comparison: equal, newer, older, `v`-prefixed, malformed tags, pre-release tags, and numeric-not-lexical ordering (`0.10.0` > `0.9.9`).
- Throttle and opt-out gating: first run, within a day, a full day old, opted out, and a clock moved backwards.
- Notice state: a failed check advances the throttle without raising a notice, dismissal holds for the session, and a newer release re-raises it.
- Settings round-trip plus a legacy `{}` document receiving the defaults.

## Verification

- `cargo build --workspace`: clean
- `cargo clippy --workspace --all-targets`: no new warnings (2 pre-existing pointer-cast warnings in `vibez-plugin-host` remain untouched)
- `cargo test --workspace`: all pass
- `cargo fmt --check`: clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5196,6 +5196,7 @@ dependencies = [
  "dirs 5.0.1",
  "iced",
  "notify",
+ "reqwest",
  "rfd",
  "rtrb",
  "serde",
--- a/crates/vibez-ui/Cargo.toml
+++ b/crates/vibez-ui/Cargo.toml
@@ -20,6 +20,9 @@ base64 = { workspace = true }
 serde_json = { workspace = true }
 notify = { workspace = true }
 rfd = "0.15"
+# Same client and TLS backend the Dropbox integration already links, so
+# the release check adds no new transitive stack.
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "json"] }
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "net", "time", "sync"] }
 x11rb = "0.13"
 dirs = "5"
--- a/crates/vibez-ui/src/app/mod.rs
+++ b/crates/vibez-ui/src/app/mod.rs
@@ -192,6 +192,7 @@ mod views_perform_record;
 mod views_perform_sections;
 mod views_settings;
 mod views_settings_appearance;
+mod views_settings_dropbox;
 mod views_settings_perform;
 mod views_shell;
 mod views_transport;
--- a/crates/vibez-ui/src/app/mod.rs
+++ b/crates/vibez-ui/src/app/mod.rs
@@ -194,6 +194,7 @@ mod views_settings;
 mod views_settings_appearance;
 mod views_settings_dropbox;
 mod views_settings_perform;
+mod views_settings_updates;
 mod views_shell;
 mod views_transport;
 mod window_policy;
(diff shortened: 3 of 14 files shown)


alexanderwanyoike:
## Review

**Approve.** Independently verified on the branch: 463 vibez-ui tests green (20 new), clippy zero new warnings; diff read in full.

This is the strongest of the four quick-win PRs. What stands out:

- **The policy is all pure functions** (`should_check`, `newer_release`, `record_result`) with the clock and network injected at the edges, which is why edge cases this feature will never be able to reproduce on demand are still tested: clock moved backwards treats the check as due rather than locking out until real time catches up, a failed check advances the throttle so an offline studio machine backs off instead of retrying every launch, and a release newer than a dismissed one re-raises the notice while re-reporting the same one stays dismissed.
- **The prerelease guard is a product decision done right**: a stable user is never nudged toward an rc, and an unparseable tag is treated as a tagging scheme this build predates, where silence is the safe answer. Both reasons are in the doc comment.
- **Failure philosophy is exactly right for a DAW**: every network failure is indistinguishable and silent, so an unreachable network behaves identically to the feature being off. Machines on stages and in studios thank you.
- **Deliberate conflict avoidance**: its own opener call instead of extending `async_helpers.rs` (where #134 adds one), a new Updates tab instead of Appearance (where #136 works), and splitting the Dropbox tab rather than the Appearance tab. All three choices were made with the concurrent branches in mind and stated.
- The chip yields the status bar to an in-flight export, and opening the releases page retires the notice through the same dismissal path.

Two observations, neither blocking:

1. **Re-enabling the toggle does not check until the next launch.** `should_check` only runs at startup, so a user who flips the setting on sits in the throttle window until tomorrow or a restart. Acceptable for a passive feature; a follow-up could fire an immediate check when the toggle transitions to on and the throttle allows.
2. `UpdateCheckState` lives in `AppState` while #136 put `interface_scale` on the `App` shell. Both placements are argued and defensible (the notice is rendered state; the scale is not), just noting the split so a future consolidation of "app preferences at runtime" knows both homes exist.

Merge-queue note: the PR body's conflict outlook against #136 (both split `views_settings.rs`, different regions) matches my reading; expect small mechanical conflicts in the tab bar and match arm only.

PR #144, opened 2026-08-09
feat(perform): mute tracks at section end

alexanderwanyoike:
## Summary
- add End of Section to the Track Mute quantization picker
- resolve the mute against the active Section's next wrap/end boundary in the audio engine
- apply immediately when stopped or when no Section is active
- preserve existing serialized Immediate, 1 Beat, and 1 Bar settings

## Test plan
- `cargo fmt --check`
- `cargo test -p vibez-core track_mute_quantization_tests`
- `cargo test -p vibez-engine engine::mute_event_tests`
- `cargo test -p vibez-engine end_of_section_track_mute_uses_the_active_section_wrap`
- `cargo test -p vibez-ui track_mute_quantization`
- `cargo clippy -p vibez-core -p vibez-engine -p vibez-ui --all-targets -- -D warnings`

### DIFF
--- a/crates/vibez-engine/src/engine_automation_commands.rs
+++ b/crates/vibez-engine/src/engine_automation_commands.rs
@@ -136,6 +136,7 @@ impl AudioEngine {
         track.queued_mute = Some(QueuedTrackMute {
             muted,
             effective_at_samples,
+            end_of_section: quantization == TrackMuteQuantization::EndOfSection,
         });
         let _ = self.event_tx.push(EngineEvent::TrackMuteQueued {
             track_id,
@@ -159,6 +160,17 @@ impl AudioEngine {
                 .map(|queued| (track.id, queued))
         }) {
             self.apply_track_mute_at(track_id, queued.muted, queued.effective_at_samples);
+        }
+    }
+
+    pub(super) fn apply_end_of_section_track_mutes(&mut self, at_samples: u64) {
+        while let Some((track_id, muted)) = self.tracks.iter().find_map(|track| {
+            track
+                .queued_mute
+                .filter(|queued| queued.end_of_section)
+                .map(|queued| (track.id, queued.muted))
+        }) {
+            self.apply_track_mute_at(track_id, muted, at_samples);
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/vibez-engine/src/engine_automation_commands.rs
```diff
@@ -106,10 +106,21 @@ impl AudioEngine {
...
-        let beats = quantization
-            .beats()
-            .expect("Immediate Track Mutes return before boundary resolution");
-        let effective_at_samples = self.next_grid_boundary(now, beats);
+        let effective_at_samples = if let Some(boundary) = quantization.musical_boundary() {
+            boundary
+                .beats()
+                .map_or(now, |beats| self.next_grid_boundary(now, beats))
+        } else {
+            self.active_section
```
alexanderwanyoike, line 114: **End of Section mute fires at a stale time if the section is switched before its wrap.**

The mute is resolved here to an absolute sample time from the active section's natural wrap at queue time, but a queued section launch can end that section earlier and nothing reconciles the pending mute: `process_section_multitrack` swaps `active_section` at the queued boundary without touching `queued_mute` (`cancel_queued_track_mutes` only runs on natural non-loop section end and on stop).

Repro: section A (8 bars, looping) playing at bar 2; queue a track mute with End of Section (effective_at = A's wrap, 6 bars away); launch section B quantized to 1 Bar. The mute fires 5 bars into section B at a musically meaningless moment instead of a section boundary.

Options: cancel/requeue pending End of Section mutes in `activate_section`, or store the queued mute as section-relative rather than an absolute sample.

alexanderwanyoike (commented):
Reviewed. One finding, inline below. Otherwise solid: the new enum keeps the exact serde names of the old MusicalBoundary alias so persisted ui.json settings stay compatible, and the stopped/no-section fallbacks match the PR intent.

crates/vibez-engine/src/engine_automation_commands.rs
alexanderwanyoike: Fixed in ae48240. Queued mutes now retain End-of-Section intent, and an earlier Section activation applies them at that exact transition boundary. Added end_of_section_track_mute_follows_an_earlier_section_transition; the focused Section-mute and mute-event suites plus clippy pass.

PR #145, opened 2026-08-09
feat(arrange): trim clips by captured track mutes

alexanderwanyoike:
## Summary
- add **Trim Track Mutes** to the Edit menu and Arrange clip context menu
- replace selected audio and MIDI clips with only their unmuted captured spans
- preserve looped audio source phase and materialize/rebase looped MIDI fragments
- keep the resulting fragments selected so Ctrl+A supports whole-arrangement trimming

## Verification

### DIFF
--- a/crates/vibez-ui/src/domains/arrangement/media_ops.rs
+++ b/crates/vibez-ui/src/domains/arrangement/media_ops.rs
@@ -6,6 +6,7 @@ use std::collections::HashSet;
 use crate::message::{AudioQuantizeSuccess, AutoWarpOutcome, ClipWarpSuccess};
 use std::sync::Arc;
 
+use vibez_core::automation::{AutomationLane, AutomationTarget};
 use vibez_core::id::{ClipId, TrackId};
 use vibez_core::midi::MidiNote;
 use vibez_engine::commands::EngineCommand;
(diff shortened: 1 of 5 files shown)


crates/vibez-ui/src/domains/arrangement/media_ops.rs
```diff
@@ -61,7 +62,259 @@ fn visible_notes(clip: &UiNoteClip, local_start: f64, local_end: f64) -> Vec<Mid
     visible
 }
 
+/// Unmuted portions of `[start, end)` in project beats. Track Mute is a
+/// stepped lane and deliberately imposes no state before its first point, so
+/// uncaptured material before the first event is retained.
+fn unmuted_beat_ranges(lane: &AutomationLane, start: f64, end: f64) -> Vec<(f64, f64)> {
```
alexanderwanyoike, line 68: **Adjacent unmuted windows are not merged, so clips get split at every lane point rather than at mute-state changes.**

This emits one range per inter-point window and classifies each by its left edge, but never coalesces consecutive unmuted windows. A clip spanning beats 0-8 whose TrackMute lane has two redundant 0.0 points inside it (e.g. from manual lane edits) yields `[(0,2.5),(2.5,6),(6,8)]`, which fails the `covers_original` check, so the clip is destructively removed and replaced by three contiguous fragments plus engine RemoveClip/AddClip churn even though nothing was muted.

Fix: merge ranges where `next.0 == prev.1` before returning (or only emit boundaries where the mute state actually flips).

alexanderwanyoike (commented):
Reviewed. One finding, inline below. The semantics otherwise check out: the 0.5 threshold and the no-state-before-first-point rule exactly match the engine's stepped-lane playback in channel_strip.rs, loop phase is preserved via the same helper the split ops use, and undo plus dirty-marking come for free from is_timeline_editor_message/marks_dirty.

crates/vibez-ui/src/domains/arrangement/media_ops.rs
alexanderwanyoike: Fixed in d8b2c44. Consecutive unmuted windows are coalesced before the covers-original check, so redundant 0.0 points leave the clip and engine untouched. Added a regression for two redundant points; all four Track Mute trim tests plus clippy pass.

PR #146, opened 2026-08-09
feat(audio): show callback CPU load in toolbar

alexanderwanyoike:
## Summary
- measure audio processing time against each device buffer deadline in the real-time callback
- publish a smoothed, lock-free CPU load value to the UI
- show a compact CPU percentage between the master level meter and tempo controls
- use success/warning/danger states to make deadline pressure visible

## Verification
- `cargo test -p vibez-audio-io audio_stream --lib`
- `cargo test -p vibez-ui views_transport --lib`
- `cargo clippy -p vibez-audio-io -p vibez-ui --all-targets -- -D warnings`

### DIFF
--- a/crates/vibez-audio-io/src/audio_stream.rs
+++ b/crates/vibez-audio-io/src/audio_stream.rs
@@ -5,10 +5,10 @@
 //! inside the real-time audio callback.
 
 use std::fmt;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use std::sync::mpsc::{self, Receiver, SyncSender};
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 use cpal::{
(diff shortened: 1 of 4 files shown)


crates/vibez-audio-io/src/audio_stream.rs
```diff
@@ -366,14 +406,19 @@ impl AudioOutputStream {
...
-                        return;
+                        processed = true;
                     }
                 }
                 // Fallback: silence
-                data.fill(0.0);
+                if !processed {
+                    data.fill(0.0);
+                }
+                callback_cpu_load.record(started.elapsed(), data.len() / ch, rt_sample_rate);
```
alexanderwanyoike, line 421: **The CPU readout freezes at its last value after a stream failure.**

The SilenceAndYield early return above exits before this `record` call, and the UI only zeroes `audio_cpu_load_percent` when `_stream` is `None`. So when a device disappears and StreamHealth marks the stream failed, the toolbar keeps showing e.g. a green 45% indefinitely while no audio is processed, contradicting the failed stream-health indicator next to it.

Cheapest fix: also record (or store 0) on the SilenceAndYield path, or have `poll_audio_stream_events` zero the value when stream health is failed.

alexanderwanyoike (commented):
Reviewed. One minor finding, inline below. The measurement itself is safely lock-free (relaxed atomic in basis points, Instant::now is fine in the callback) and the silence-fallback restructuring preserves the old behavior exactly.

crates/vibez-audio-io/src/audio_stream.rs
alexanderwanyoike: Fixed in 3efb45a. CPU load is now applied after lifecycle events and forced to zero whenever stream health is not Running, so the stale atomic value cannot return on later ticks. Added the failed-stream regression; audio-stream tests and clippy pass.

PR #147, opened 2026-08-09
feat(project): add recent projects submenu

alexanderwanyoike:
## Summary
- add a Recent Projects submenu to File with filenames and parent paths
- keep the six most recent successful project opens/saves in global UI settings
- deduplicate and promote reopened projects to the top
- add Clear Recent Projects and preserve compatibility with existing settings files

## Verification
- `cargo test -p vibez-ui --lib` (492 passed)

### DIFF
--- a/crates/vibez-ui/src/app/menu_lifecycle.rs
+++ b/crates/vibez-ui/src/app/menu_lifecycle.rs
@@ -16,7 +16,10 @@ pub(super) fn dismiss(state: &mut AppState, overlay: MenuOverlay) -> bool {
 
     match overlay {
         MenuOverlay::ArrangementContext => state.view.context_menu = None,
-        MenuOverlay::File => state.project.file_menu_open = false,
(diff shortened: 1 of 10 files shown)


alexanderwanyoike:
Reviewed, no findings. Dedupe/cap/promote logic is correct, old settings files deserialize fine (serde default covers the new field), and clicking a recent entry goes through the same ProjectOpenPathSelected path as the file dialog, so no behavior regression.

One non-blocking observation: entries that fail to load (deleted/moved files) stay in the list; most DAWs prune the entry on load failure. Cheap follow-up if you want it.

alexanderwanyoike:
Implemented the suggested cleanup in d7e3dba. The attempted path now accompanies its async load result; on failure, that exact recent entry is pruned and the updated settings are persisted. Added a focused pruning regression, and all 493 vibez-ui tests plus clippy pass.

PR #149, opened 2026-08-11
Release v0.1.5

alexanderwanyoike:
Patch release for deeper Perform capture editing, project recall, responsive Remote browsing, audio visibility, and the Vibez site.

## Summary

- Add End of Section Track Mute quantization, including correct behavior when another Section transitions before the natural wrap and when a one-shot Section reaches its terminal boundary.
- Add Trim Track Mutes for selected Arrange audio and MIDI clips, preserving loop phase and avoiding fragmentation at redundant automation points.
- Show real-time audio callback CPU load beside the master meter and tempo controls, with truthful reset behavior after stream failure.
- Add a six-entry Recent Projects submenu with promotion, deduplication, clearing, automatic pruning of failed paths, and correct alignment with its parent menu item.
- Keep large saved Remote catalogs off the UI startup path while showing an honest loading state.
- Keep Dropbox refresh reconciliation, sorting, folder indexing, cache matching, persistence snapshots, and retired catalog disposal off the UI thread.
- Preserve the committed Dropbox cursor across intermediate refresh batches, rebase prepared snapshots over newer materialization/cache state, and reject stale worker failures before they can alter the active refresh.
- Feature the Vibez promotional workflow video on the GitHub Pages site using its existing visual language and responsive layout.
- Update the README and bump the Cargo workspace and lockfile from 0.1.4 to 0.1.5.

## Test plan

- [x] `cargo check --workspace`
- [x] `cargo test --workspace`
- [x] `cargo clippy --workspace --all-targets --exclude vibez-plugin-host -- -D warnings`
- [x] `cargo clippy -p vibez-plugin-host -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] Feature and fix PRs #144 through #153 passed CI across Linux, macOS, and Windows
- [x] Manual QA: Recent Projects submenu alignment
- [x] Manual QA: 383,557-entry Megalodon Dropbox catalog remains responsive during startup and refresh
- [ ] Final release QA

## After merging

Tag and push `v0.1.5` to trigger the release-artifact workflow:

```sh
git tag v0.1.5
git push origin v0.1.5
```

Prepared with Codex.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5085,7 +5085,7 @@ checksum = "0b928f33d975fc6ad9f86c8f283853ad26bdd5b10b7f1542aa2fa15e2289105a"
 
 [[package]]
 name = "vibez"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "iced",
  "libloading 0.8.9",
@@ -5095,7 +5095,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-audio-io"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "audio_thread_priority",
  "cpal",
@@ -5109,7 +5109,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-core"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "serde",
  "serde_json",
@@ -5117,7 +5117,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-dropbox"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "base64",
  "dirs 5.0.1",
@@ -5136,7 +5136,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-dsp"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "signalsmith-stretch",
  "vibez-core",
@@ -5144,7 +5144,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-engine"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "rtrb",
  "vibez-core",
@@ -5154,14 +5154,14 @@ dependencies = [
 
 [[package]]
 name = "vibez-instruments"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "vibez-core",
 ]
 
 [[package]]
 name = "vibez-plugin-host"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "clap-sys",
  "dirs 5.0.1",
@@ -5178,7 +5178,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-project"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "rusqlite",
  "serde",
@@ -5190,7 +5190,7 @@ dependencies = [
 
 [[package]]
 name = "vibez-ui"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "base64",
  "dirs 5.0.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.1.4"
+version = "0.1.5"
 edition = "2021"
 license = "GPL-3.0-or-later"
 repository = "https://github.com/vibez-daw/vibez"
--- a/README.md
+++ b/README.md
@@ -142,3 +142,7 @@ where a redundant-looking cast can be load-bearing on another target.
 ## License
 
 GPL-3.0-or-later. See [LICENSE](LICENSE).
+
+---
+
+This one's for you Mum 🥲 I miss you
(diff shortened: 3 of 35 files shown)


alexanderwanyoike:
## Code review (v0.1.5 release diff, dev → main)

Full diff reviewed against source; CI is green on all platforms.

### Bugs

1. **`crates/vibez-ui/src/app/dropbox_io.rs:162`** - The empty-changes fast path in `prepare_remote_catalog_refresh` unconditionally assigns `catalog.checkpoint = checkpoint`. At a mid-refresh save interval the caller passes `checkpoint: None`, so a 10-page interval with zero relevant changes wipes the saved Dropbox delta cursor and persists `None`. An interrupted refresh then falls back to a full relisting. The old `flush_remote_catalog_pages` guard prevented this, and the new `reconcile_remote_catalog` early-return gets it right (only overwrites when `Some`); the fast path should match.

2. **`crates/vibez-ui/src/app/update_remote.rs:233`** - `on_remote_catalog_refresh_prepared` wholesale-replaces catalog and availability with snapshots captured before the worker ran. If the user materializes a remote sample during the prepare window (long: the worker clones and re-sorts the whole catalog), the completion handler's `Cached` badge and the `derived_metadata` written via `Arc::make_mut` (update_remote.rs:494) are silently reverted, and the next persist writes the metadata-less catalog to disk. Fix direction: apply UI-thread deltas on top of the prepared result, or route materialization through the same prepared pipeline.

3. **`crates/vibez-engine/src/engine_automation_commands.rs:108`** - An End of Section mute queued during a non-looping section never applies. `effective_at_samples` lands exactly on the section-end sample, but a one-shot section starts no render segment at that sample; the section-ended path (`engine.rs:352`) calls `cancel_queued_track_mutes`, so the queued mute silently evaporates. Tests cover the looping wrap and the transition case, but not the one-shot stop.

4. **`crates/vibez-ui/src/app/update_project.rs:177`** - Any `ProjectLoaded` error prunes the path from Recent Projects, including transient failures (unmounted drive, Dropbox media unresolvable offline), and persists the pruned list. Consider pruning only on "file not found" and surfacing other errors without forgetting the entry.

5. **`crates/vibez-ui/src/app/update_remote.rs:220`** - The `Err` arm of `on_remote_catalog_refresh_prepared` has no generation guard (the `Ok` arm checks `is_current`). A failed prepare from a cancelled generation (disconnect + reconnect mid-refresh) stomps the new refresh's `catalog_state` and status text with `Stale`.

### Design / DRY

6. **`crates/vibez-core/src/perform.rs:411`** - `TrackMuteQuantization` is a verbatim copy of `SectionLaunchQuantization`: same four variants, `ALL`, `label()`, `musical_boundary()`, and `Display`, differing only in `#[default]`. One shared enum with the default at the settings field halves it.

7. **`crates/vibez-ui/src/app/dropbox_io.rs:15`** - `run_remote_startup_loader` and `run_remote_refresh_preparer` are byte-identical apart from the error string, each with a near-identical test; every other spawn_blocking wrapper lives in `async_helpers.rs`. Related: the startup task rebuilds the cached-revision rule inline (dropbox_io.rs:47) instead of calling `refreshed_remote_availability`, and the copies already differ (the startup copy does not skip folders).

8. **`crates/vibez-ui/src/domains/arrangement/media_ops.rs:227`** - The trim op re-implements the remove-original / `AddClip`(nine fields) / retain / extend sequence that `op_split_audio_clip` and `op_split_note_clip` already carry; third copy of a command list that must be touched when a clip field is added (as `groove_grid` was). Extract shared `replace_audio_clip` / `replace_note_clip` helpers.

9. **`crates/vibez-ui/src/app/views_overlays.rs`** - Crosses the 1,000-line ceiling for the first time (901 → 1,056); the recent-projects card is the natural split. Also `recent_btn` (line 600) is the one menu row without `.height(FILE_MENU_ITEM_HEIGHT)`, so the row the submenu aligns to renders taller than the 32px the arithmetic assumes, and `RECENT_PROJECTS_MENU_ITEM_INDEX = 2` silently breaks on reorder (the test checks the arithmetic, not the correspondence).

### Nits

- `audio_stream.rs:55`: `previous == 0` doubles as the uninitialized sentinel, so an idle callback re-seeds the EMA; the smoothed value also survives `reconfigure()`, briefly blending old and new buffer deadlines.
- `update_remote.rs:243`: raw `std::thread::spawn` per prepared refresh (plus one per stale generation) just to run destructors; `tokio::task::spawn_blocking` is the established pool, and the empty-changes case has nothing worth deferring.
- `views_transport.rs:63`: `audio_cpu_indicator` duplicates the pill chrome of `audio_stream_indicator` (third and fourth copies in the perform views); one `status_pill()` helper.
- Latent, pre-existing: `queue_track_mute` stores via `channel_mut` (resolves buses/master) but all appliers scan `self.tracks` only. Unreachable from the current UI; worth a debug_assert.

Verified clean: serde back-compat for `TrackMuteQuantization` and `recent_project_paths`, all new messages routed, trim in the undoable list, trim math (loop phase, note materialization, boundary equality), and Recents opening consistent with the guardless regular Open flow.

**Verdict:** fine to ship since dev is already carrying this and CI is green; file issues for bugs 1-3 and fix in a patch, DRY items as follow-up cards, one PR each.

PR #163, opened 2026-08-14
feat: add MIDI velocity editing lane

alexanderwanyoike:
## Summary

- add a time-aligned velocity lane beneath the piano roll
- edit one note or a selected group with relative, clamped drag behavior
- preserve shared note selection and group each drag into one undo gesture
- keep the MIDI detail editor usable with a velocity-aware minimum panel height
- synchronize velocity edits with the audio engine and project persistence

## Verification

- `cargo test --workspace`
- `cargo clippy --workspace --all-targets --exclude vibez-plugin-host -- -D warnings`
- native UI QA: selection, drag to 127, undo, redo, autosave, and reopen persistence

## Release slice

Production Depth 01 — MIDI velocity lane.

### DIFF
--- a/crates/vibez-ui/src/app/views_detail.rs
+++ b/crates/vibez-ui/src/app/views_detail.rs
@@ -18,11 +18,12 @@ use crate::message::Message;
 use crate::state::{ArrangementSelection, DetailPanelTab, TimelineEditorState, UiClip};
 use crate::theme as th;
 use crate::widgets::audio_clip_detail::AudioClipDetailWidget;
-use crate::widgets::piano_roll::PianoRollWidget;
+use crate::widgets::piano_roll::{PianoRollWidget, VelocityLaneWidget};
 
 use super::*;
 
 const DETAIL_PANEL_MIN_HEIGHT: f32 = 180.0;
+const MIDI_DETAIL_PANEL_MIN_HEIGHT: f32 = 360.0;
 const SHELL_AND_WORKSPACE_MIN_HEIGHT: f32 = 360.0;
 const STATUS_BAR_HEIGHT: f32 = 24.0;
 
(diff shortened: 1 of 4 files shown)


alexanderwanyoike:

### Bugs

1. **fix** `crates/vibez-ui/src/widgets/piano_roll/velocity_lane.rs:372` — The drag's no-op guard compares freshly computed velocities against `drag.original_velocities`, but the model holds the *last sent* values, so a drag that returns to its start point leaves a residual edit. Scenario: drag up (messages fire), drag back to the starting pixel; every position rounding to delta 0 is skipped, so the model keeps the last ±1-step value while the cursor is back at the origin. The user releases believing nothing changed and the note is one velocity step off. Fix: track the last-sent vector in `VelocityDrag` and compare against that (or always send; the message is idempotent).

### Design/DRY

2. **fix** `crates/vibez-ui/src/app/views_detail.rs:239` — The two-line `midi_clip_editor_visible` predicate is duplicated between `view_detail_panel` and `detail_panel_drag_height`. If one site changes, the drag preview and the rendered panel height silently disagree. Extract a tiny `fn midi_clip_editor_visible(&self) -> bool` on `App` and call it from both.

### Nits

3. **fix** `crates/vibez-ui/src/widgets/piano_roll/velocity_lane.rs:421` — `requested_delta.clamp(1 - min_velocity, 127 - max_velocity)` panics if a selection ever contains stored velocities spanning more than 126 (a 0 next to a 127). Unreachable today (MIDI input normalises velocity-0 to note-off; every editor path clamps to 1..127), but the invariant lives far from this code. Clamp each captured velocity to `1..=127` when building `original_velocities` so the panic is structurally impossible.
4. **note** — A drag emits one `EditNote` engine command per selected note per mouse-move; select-all on a dense clip means hundreds of commands per frame. Nothing is dropped (the command queue spills to its pending buffer), so acceptable until profiling says otherwise; the fix then is coalescing during the drag.
5. **note** — `widgets/piano_roll/mod.rs` was already over the 1,000-line ceiling (1,023 → 1,026). The three added lines are unavoidable mod declarations, but the file is now on the split-me list; the inline test module is the natural seam.

Also verified, no action needed: `SetNoteVelocities` is automatically dirty-marking and undoable (`marks_dirty` is an exclusion list), the per-drag undo gesture coalescing is pinned by tests, out-of-range note indices are ignored with a test, and the relative-drag group clamp semantics are tested.

alexanderwanyoike:
Addressed the three actionable mini-review findings in the latest commit:

- track the last emitted velocity vector, so returning to the drag origin sends the restoring edit
- add a regression that reproduced the missing restore before the fix and now passes
- normalize captured stored velocities to `1..=127` before group clamping, with a hardening test
- centralize `midi_clip_editor_visible()` for render and drag-height calculations

Local verification: focused return-to-origin regression, stored-velocity hardening regression, 24 piano-roll tests, 5 detail-view tests, strict UI Clippy, and formatting all pass.

PR #165, opened 2026-08-14
feat(audio): add persistent soundcard settings

alexanderwanyoike:
## Outcome

Adds a production-ready Settings → Audio hardware surface:

- enumerate and select Audio Input and Audio Output devices
- choose only supported Sample Rates and Buffer Sizes
- persist hardware choices outside project files
- apply output changes transactionally without losing the working stream on failure
- fall back to System Default while retaining a missing named-device preference
- rescan and reconnect from explicit UI state
- support CPAL float and integer output formats through an isolated callback adapter

## Verification

- `cargo test --workspace`
- `cargo clippy --workspace --all-targets --exclude vibez-plugin-host -- -D warnings`
- `git diff --check`

Native PipeWire/ALSA QA used an isolated UI config. Vibez switched live from `default` to `pulse`, changed from 44.1 kHz / 512 frames to 48 kHz / 128 frames, remained an active PipeWire stream, and restored the configuration after restart. A simulated missing USB interface preserved the named preference, ran on System Default, and exposed reconnect status.


alexanderwanyoike:
## mini-code-review — verdict: merge after fixing findings 1 and 2

Focus requested: bugs, DRY, architecture adherence. Architecture verdict up front: the split is right (pure choice/capability logic in `domains/audio_settings.rs` with tests, App actions in `app/audio_settings.rs`), the ui.json persistence pattern is followed exactly (serde default fn, `Default` impl, `persist_ui_settings`, `App::new` load, empty-JSON back-compat test), and hardware config is correctly app-scoped, never project state. Build-before-pause reconfigure and non-f32 sample-format support are real robustness wins.

### Bugs

1. **fix** `crates/vibez-ui/src/state/mod.rs:597` — A rejected configuration leaves `audio_stream_health` stuck at `Rebuilding` forever. Sequence: `reconfigure` emits `Rebuilding` (arm at state/mod.rs:591 sets health = Rebuilding), then `ConfigurationRejected`, whose arm sets only status text and no later event ever fires (the previous stream is still live by definition of the event). Settings shows "Applying configuration…" indefinitely and the CPU meter reads 0 because `update_audio_cpu_load` gates on `Running`. The new unit test misses this by starting from the default `Running` state instead of replaying the `Rebuilding` → `ConfigurationRejected` sequence. Fix: the `ConfigurationRejected` arm should restore `AudioStreamHealth::Running`; update the test to replay both events.

2. **fix** `crates/vibez-ui/src/app/audio_settings.rs:24` — The startup fallback only drops the device name and keeps the saved sample rate and buffer. If the saved rate/buffer combo is the unsupportable part (dock unplugged, new default device with narrower caps), the fallback also fails and startup is silent even though System Default at device defaults would work; when `preferred_output_name` is `None` there is no fallback at all. Degrade progressively: saved config → saved minus device name → device defaults (no rate/buffer) before giving up.

### Design/DRY

3. **fix** `crates/vibez-audio-io/src/audio_host.rs:207` — `output_device_info` and `input_device_info` are near-identical twins differing only in which cpal accessors they call. One helper taking the default-config and supported-configs results (or a direction parameter) halves it.

4. **fix** `crates/vibez-ui/src/domains/audio_settings.rs:240` — `output_devices_with_configs()` deep-clones the whole device list (nested config vecs included) on every call, and both `output_choices()` and `selected_output_choice()` call it on every render frame while the settings modal is open; the default-availability expression is also duplicated between those two. Return references (or filter inline) and extract a `default_output_available()` helper.

### Nits

5. **note** — On a successful change the handler writes the informative "Audio Output running — device, rate, frames" status, then the queued `Rebuilding`/`Recovered` events overwrite it with "Rebuilding audio stream…" and "Audio stream recovered" on the following ticks. Suppressing the generic statuses for user-initiated changes would keep the useful one visible.
6. **note** `crates/vibez-audio-io/src/audio_stream/callback.rs:216` — `scratch.resize(data.len(), 0.0)` can allocate on the audio thread when a backend ignores the fixed-size hint (at most once per new size; the f32 fast path avoids it). Given the allocation-free audio-thread rule, consider preallocating to the device's reported max buffer range where available.
7. **note** `crates/vibez-ui/src/app/audio_settings.rs:150` — Selecting an input persists the preference and reports "Audio Input ready", but nothing opens an input stream until recording lands; soften the wording so it does not oversell.
8. **note** — `views_settings.rs` is now 933 lines; split the audio tab out before the file crosses the 1,000-line ceiling.

Verified and dropped, no action needed: mono-only devices are safe (the engine guards `channels >= 2` on every pan write, so removing the old stereo clamp is correct), and the brief reconfigure overlap window is mutex-protected and a documented trade-off.

alexanderwanyoike:
Resolved all eight review findings in d598e91:

1. ConfigurationRejected now restores Running after the real Rebuilding → Rejected sequence; regression test added.
2. Startup now degrades through saved config → System Default with saved rate/buffer → System Default device defaults, including when no named device was saved; policy tests added.
3. Input/output host enumeration now shares one direction-aware mapper.
4. Output choices iterate borrowed usable devices and share default-availability policy; no deep per-frame device clones.
5. User-initiated lifecycle events are drained synchronously so the informative success status remains visible.
6. Integer/float64 conversion scratch is fully preallocated off the audio thread; oversized callbacks silence instead of allocating.
7. Audio Input copy now says selected and makes clear recording starts only when a track is armed.
8. The Audio tab moved into views_settings_audio.rs; the settings shell is down from 933 to 441 lines.

Local verification: formatting, strict workspace Clippy, cargo test --workspace, native launch, active PipeWire stream.

PR #166, opened 2026-08-15
feat: record hardware audio input to Arrange

alexanderwanyoike:
## Outcome

Audio Tracks can now arm one hardware-input target, choose mono/stereo routing and Off/Auto/On monitoring, see the pre-input meter, and record a project-owned Audio Clip from the current Arrange cursor.

## What changed

- persists per-Audio-Track input route and monitoring mode with legacy defaults
- opens the selected CPAL input on demand and converts every supported sample format
- bridges input to the output clock through bounded lock-free SPSC rings
- injects monitored input before the target Track's effects and mixer path
- latches the exact record start on the output callback and acknowledges Stop before final draining
- encodes WAV off-thread, stages Project Media first, and lands the Clip as one Undo transaction
- abandons the take cleanly on device loss, unsupported routing, or buffer overflow
- documents the hardware-input real-time and persistence boundaries

## Verification

- `cargo test --workspace`
- focused bridge, routing, monitoring, engine-target, lifecycle, Project Media, and save/reopen tests
- native PipeWire QA: armed a physical input, observed the pre-input meter, recorded an 8.20 s non-silent take, landed its waveform, and removed it with one Undo
- native save/reopen QA: a recorded take reopened from embedded Project Media at the same placement

## Scope

This intentionally supports one hardware-input Audio Track at a time. MIDI Track resampling, Section recording, punch/comping, latency compensation, and Clip editing remain separate release slices.


alexanderwanyoike:
## mini-code-review (2 agents) — verdict: needs another pass

Two-agent review (lead + independent second pass), every finding verified against `origin/feat/audio-input-recording` source. Architecture is right: domain state in `domains/audio_recording.rs`, app actions in `app/audio_recording.rs`, route/monitoring as project state with hardware app-scoped, correct SPSC ring ordering, allocation-free audio paths, one-undo-step transaction. The recording lifecycle, however, has holes on primary workflows.

### Bugs

1. **fix** `crates/vibez-ui/src/app/update.rs:104` — Spacebar does not stop recording. The intercept matches only `TransportMsg::Stop`, but space sends `TogglePlayback` (keyboard.rs:322), which converts to Stop inside the transport domain (domains/transport.rs:101-107), bypassing the intercept. Transport stops and seeks to 0 while the bridge keeps capturing; the finalized take contains all post-space audio appended from the latch position. Intercept TogglePlayback-while-playing as a stop as well.
2. **fix** `crates/vibez-audio-io/src/audio_input.rs:286` — No input/output clock-drift handling. Slower input: `self.input.pop().unwrap_or([0.0; 2])` silently interleaves zero frames into monitoring and the recorded take; there is no underrun counterpart to `report_overflow`, so the take is corrupted with no warning. Faster input: ring occupancy grows until the 262,144-frame ring overflows, destroying the take (see 7). Minimum v1 fix: count and surface underrun-substituted frames; resampling can be a follow-up card.
3. **fix** `crates/vibez-ui/src/app/audio_recording.rs` (`poll_audio_input`) + `audio_input.rs:302` — Output-stream death during recording wedges the app. `recording_stopped` is set only by the output callback; if the output device dies mid-take, phase `Stopping` never resolves and `is_busy()` permanently blocks arm/route/monitoring/hardware changes until project close. Resolve the stop when `audio_stream_health` becomes Error, or add a timeout.
4. **fix** `crates/vibez-ui/src/app/update.rs` — `Seek`, `SeekToBeat`, and `ToggleArrangementLoop` are not blocked during recording (only loop-at-start is checked). A ruler click mid-take seeks the engine under a contiguous capture, so post-seek audio lands at positions it does not correspond to. Block them while `is_capturing()`.
5. **fix** `crates/vibez-ui/src/app/audio_recording.rs` (`finish_audio_recording`) — The async finalize result can land in the wrong project. `close_project` resets recording state but cannot cancel the in-flight `Task::perform`; the handler then `AddClip`s to a dead track id, `arrange_content_mut` (`ensure(id)`, state/mod.rs:745) creates timeline content for a ghost track, and an undo snapshot + dirty flag hit the new project. Guard on the target track still existing.
6. **fix** `crates/vibez-ui/src/app/audio_recording.rs` (`sync_audio_input_runtime`) — Switching the input picker back to System Default never reopens the stream: `must_reopen` uses `expected_name.is_some_and(...)`, so with `preferred_input_name = None` the previously named device silently stays in use until a sample-rate change or rearm. Wrong-device capture.

### Design / policy

7. **fix** `crates/vibez-ui/src/app/audio_recording.rs` — Overflow during capture clears `captured_frames` and abandons the transaction: total loss of the take. The already-drained frames are valid; salvage them as a truncated clip with a warning instead of discarding a long performance over a transient buffer condition.
8. **note** — No input-latency compensation: clips land late by roughly input+output latency (10-40 ms typical), and each engine `try_lock` miss adds permanent un-flagged ring occupancy. Acceptable as a documented v1 limitation, but ARCHITECTURE.md's "exact Arrange sample position" describes the clock, not the audio; document the gap and card the round-trip compensation.

### DRY

9. **fix** `crates/vibez-audio-io/src/audio_input.rs:417` — `select_input_config` is a near-line-for-line clone of `select_output_config` (audio_stream.rs), same filters/sort/fallback/error format; the 11-arm SampleFormat dispatch and the error enum also mirror their output counterparts. Fold the config selection (direction parameter, as audio_host.rs already does after #165).
10. **fix** — The "find the persisted monitoring-On track" search exists three times (`persisted_monitor_on_track` in audio_recording.rs, inlined in project_io.rs and project_replay.rs); the input channel-count lookup twice (views_arrangement.rs, audio_recording.rs). Call the existing helper; move the channel lookup onto `AudioSettingsState`.

### Nits

11. **note** — Long-take memory: `captured_frames` grows ~1.4 GB/hour (48 kHz stereo f32) and finalize transiently ~triples it, including writing a temp WAV then `fs::read`ing it back solely to hand bytes to staging. Encode to memory or stream into the staging path.
12. **fix** `crates/vibez-audio-io/src/audio_input.rs:350` — Input peak meters overwrite per input callback with no max-hold; polled at 60 fps against 2-5 ms callbacks, the UI misses most transients and the meter under-reads during level-setting. Accumulate with max and reset on read.

Verified clean by both agents independently: StereoRing SPSC memory ordering; no double input-ring consumer during output reconfigure (old stream paused before replacement plays); `recording_stopped` release/acquire pairing guarantees the final drain sees all frames; TrackId-0 sentinel safety; allocation- and lock-free audio hot paths; engine injection ordering in both render paths (clear-before-sum, early-continue moved below injection); stop-intercept self-recursion terminates.

alexanderwanyoike:
Second polish pass pushed in `22319b9`. All review findings are addressed:

1. Space/Toggle Playback now follows the recording Stop/finalize path; native QA confirms it lands the take and resets transport.
2. Input-clock underruns are counted and surfaced as silence-substituted frames.
3. Output error and a bounded Stop-ack timeout abort cleanly instead of wedging `Stopping`.
4. Seek, ruler seek, loop changes, loop-region changes, and tempo commits are locked during capture.
5. Finalizer results require the same active finalizing session and a live target Track, so project replacement cannot create ghost content.
6. System Default resolves to its current concrete device name and reopens when that differs from the live stream.
7. Overflow stops capture and salvages the already-valid prefix with an explicit warning instead of destroying the take.
8. Architecture docs now distinguish exact output-clock placement from uncalibrated acoustic latency and retain round-trip compensation/drift correction as later recording depth.
9. Input/output config selection, stream-open errors, and the eleven-format dispatch now share one implementation.
10. Monitoring-On lookup and input channel-count lookup each have one owner/helper.
11. Completed recordings encode directly to in-memory WAV bytes for staging; the documented remaining long-take RAM scaling is explicit.
12. Input meters now max-hold atomically until the 60 fps UI read, then reset.

UI polish is included: the arm button uses the record glyph, active/inactive states are clearer, route/monitor selectors are compact themed controls, dropdown menus match Vibez surfaces, and the input meter changes to recording/monitor colors.

Verification: strict workspace Clippy, complete workspace tests (including 533 UI tests), focused 10-test recording regression set, and native PipeWire physical-input/Space-stop/UI QA all pass locally. CI is running on the pushed revision.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is

Re-reviewed 22319b9 against all 12 findings: every fix verified in source with tests (spacebar/TogglePlayback guard incl. the Stopping re-entrancy case, underrun telemetry, triple-layered output-death handling with a 2 s stop-ack timeout, seek/loop/tempo blocked mid-take, double-guarded finalizer, catalog-resolved System Default reopen, overflow salvage policy, shared stream_config + dispatch_sample_format fold, helper dedup, in-memory WAV encode, max-hold meters). ARCHITECTURE.md now documents the v1 latency/drift limitations honestly.

One note, no action needed: the unified select_stream_config tiebreaker now prefers fewer channels where the old input path preferred more; only reachable when a device has no config matching its default channel count, and route_fits_channels catches a too-narrow result with a clear error.

CI green on all platforms.

PR #167, opened 2026-08-15
feat: resample MIDI track output into audio

alexanderwanyoike:
## Summary

- add a persisted Resample input route to Audio Tracks, sourced from eligible MIDI/instrument Project Tracks
- capture the source post-instrument, post-inserts, post-mute and post-fader/pan, while excluding sends, returns, master processing, other Tracks and Browser Audition
- reuse the hardware-recording output-clock lifecycle, bounded bridge, Project Media staging, Clip insertion and one-step Undo without opening a hardware input stream
- add compact `RS · Track` routing UI, invalid/source-removal recovery, documentation and regression coverage

## Verification

- `git diff --check`
- `cargo test --workspace`

## Native QA

Created an Audio Track and instrument MIDI Track, selected `RS · MIDI 2`, armed with the recording glyph, recorded a non-silent take, stopped with Space, confirmed playback, undid/redid the take, saved, restarted Vibez and reopened with both the route and project-owned audio intact. A second take completed with `Resampled MIDI 2 · 1.11 s · one undo step`.

## Scope

Live Arrange resampling only. Audio Track/bus/master/Browser Audition/Perform sources, multi-source recording and offline Bounce/Freeze remain out of scope.

### DIFF
--- a/crates/vibez-audio-io/src/audio_input.rs
+++ b/crates/vibez-audio-io/src/audio_input.rs
@@ -128,6 +128,7 @@ pub struct AudioInputBridge {
     route: AtomicU32,
     target_track: AtomicU64,
     monitoring: AtomicBool,
+    resample_source_track: AtomicU64,
     recording: AtomicBool,
     recording_stopped: AtomicBool,
     record_start_position: AtomicU64,
@@ -151,6 +152,7 @@ impl AudioInputBridge {
             route: AtomicU32::new(0),
             target_track: AtomicU64::new(0),
             monitoring: AtomicBool::new(false),
+            resample_source_track: AtomicU64::new(0),
             recording: AtomicBool::new(false),
             recording_stopped: AtomicBool::new(true),
             record_start_position: AtomicU64::new(u64::MAX),
@@ -165,6 +167,7 @@ impl AudioInputBridge {
         let encoded = match route {
             AudioInputRoute::Mono { channel } => u32::from(channel),
             AudioInputRoute::Stereo { left } => STEREO_ROUTE_BIT | u32::from(left),
+            AudioInputRoute::Resample { .. } => return,
         };
         self.route.store(encoded, Ordering::Release);
     }
@@ -194,6 +197,20 @@ impl AudioInputBridge {
 
     pub fn target_track_raw(&self) -> Option<u64> {
         match self.target_track.load(Ordering::Acquire) {
+            0 => None,
+            raw => Some(raw),
+        }
+    }
+
+    pub fn set_resample_source(&self, source: Option<TrackId>) {
+        self.resample_source_track
+            .store(source.map_or(0, TrackId::raw), Ordering::Release);
+        self.peak_l.store(0, Ordering::Relaxed);
+        self.peak_r.store(0, Ordering::Relaxed);
+    }
+
+    pub fn resample_source_track_raw(&self) -> Option<u64> {
+        match self.resample_source_track.load(Ordering::Acquire) {
             0 => None,
             raw => Some(raw),
         }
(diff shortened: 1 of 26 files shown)


alexanderwanyoike:

No bugs found. Verified against source (not just the diff): ring gating between `clock_output` and `capture_track_output` including a stop landing mid-callback (no false stop-acks, no lost frames, Release/Acquire pairing holds); underrun counting correctly suppressed in resample mode and the hardware drift warning dropped for track-output takes; capture scratch zeroing and segment slicing in lockstep with output slices through loop-wrap, segment, and idle paths; feedback edges structurally impossible (sources are MIDI tracks, targets audio); source removal mid-take aborts cleanly with domain reset + test; resample takes are sample-accurate (no hardware latency in the path). Engine test coverage is excellent (post-fader, sends/returns/master/audition excluded, solo-silence, automation-following).

### Design/DRY

1. **fix** `crates/vibez-engine/src/engine_render.rs:67` — The loop-wrap branch is fully duplicated: the capture-aware copy (two `render_multitrack_segment` calls, `apply_track_mutes_due`, note flushes, `split_wrap_handled`) sits above the retained None-capture original, including a duplicated load-bearing comment about boundary work. The wrap boundary is exactly the code where the v0.1.5 review found a mute bug; it must not exist twice. Fold by mapping the optional capture into optional split halves (e.g. build `Option<TrackOutputCapture>` per half) and keep one branch.
2. **fix** `crates/vibez-engine/src/engine.rs:200` — Four public process variants (`process`, `process_with_live_input`, `process_with_track_output_capture`, `process_with_live_input_and_track_output_capture`) wrap one private `process_block`, and the output callback pattern-matches a 3-tuple to pick among them; a third optional source doubles the surface to eight. Expose one public entry taking an options/params struct (the codebase already uses this shape with `MultitrackRenderBlock`), collapsing both the engine API and the callback dispatch.
3. **note** — The 7-line capture-write block is duplicated between `render_multitrack_frames` and `render_idle_instruments`, extending those loops' pre-existing near-duplication; fold alongside finding 1 if convenient.

### Nits

4. **note** — Forward-compat: a project saved with a `Resample` route fails to open in builds before this PR (unknown serde variant fails the whole `TrackInfo` deserialize). No forward-compat policy exists, so informational only, but relevant while dogfooding across versions.

alexanderwanyoike:
Addressed the review in `9ab2bb6`:

- folded capture/no-capture loop-wrap rendering into one boundary path, including the single mute-boundary application, note flush and `split_wrap_handled` assignment
- replaced the four public process variants and callback tuple dispatch with one public `AudioProcessBlock` builder + `process_block` entry point
- extracted the duplicated capture sample write
- added `track_output_capture_stays_aligned_across_arrangement_loop_wrap` to lock the load-bearing boundary behavior

Verification passed: `cargo fmt --all -- --check`, `git diff --check`, strict workspace Clippy, and the complete workspace test suite. The forward-compat observation remains informational because the project has no downgrade/forward-compat policy.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is

Re-reviewed 9ab2bb6: finding 1 fixed the right way (split_track_output_capture collapses the wrap branch back to one copy, plus a new test pinning capture/output alignment across the loop wrap), finding 2 fixed via the AudioProcessBlock builder (four public variants deleted, callback dispatch reduced to two conditional builder calls), finding 3 folded into write_track_output_sample. Verified the (None, None) split path is behaviorally identical to the old no-capture branch and the builder adds zero audio-thread cost.

One tiny leftover, no action required for merge: the five-step pipeline doc comment now sits on the #[cfg(test)] process() wrapper; it belongs on process_block. Move it next time the file is touched.

CI green.

alexanderwanyoike:
Release-blocker follow-up is fixed in `2067e29`.

Root causes:
- the recording bridge was draining captured frames at UI cadence, but Arrange had no transient Audio Clip model/draw path, so there was nothing to render until Stop committed the Clip;
- normal Arrange playback still owned the existing content-end boundary, so a Record command starting at that boundary could auto-stop on the first audio callback.

What changed:
- the armed Audio Track now renders a red, display-only waveform preview from incrementally aggregated captured frames; it is bounded to 4096 peak buckets and progressively compacts for long takes;
- Audio Track Recording explicitly suspends the ordinary Arrange content-end boundary before Play and Stop restores the canonical boundary;
- the callback-latched start position is shared by the preview and final Clip.

Native QA reproduced the exact edge case: existing content ended at 8.00 s, recording began at 6.55 s, the playhead advanced through 9.76 s while the waveform visibly grew, and Space committed the same take as a normal resample Clip.

Verification:
- `git diff --check`
- `cargo test --workspace` (42 Audio I/O, 73 Core, 222 Engine, 538 UI tests plus integration/doc suites)

PR #169, opened 2026-08-15
fix: stop CPU monitor toolbar jitter

alexanderwanyoike:
## Outcome

The live Audio CPU value can cross one-, two-, and three-digit boundaries without moving any neighbouring transport controls.

## Root cause

The tooltip control wrapped a shrink-to-content container. Every additional numeral widened the entire badge and reflowed the transport row. The percentage also had no stable alignment slot.

## Fix

- Give the complete CPU badge a fixed 62 logical-pixel width.
- Put the percentage in a right-aligned 28-pixel IBM Plex Mono slot.
- Preserve the existing deadline colours and tooltip.

## Regression evidence

`cpu_indicator_keeps_one_fixed_toolbar_width_for_every_digit_count` inspects the production widget's declared width at `5%`, `13%`, `100%`, and the clamped `999%` maximum.

## Verification

- `git diff --check`

### DIFF
--- a/crates/vibez-ui/src/app/views_transport.rs
+++ b/crates/vibez-ui/src/app/views_transport.rs
@@ -19,6 +19,9 @@ use crate::widgets::vu_meter::VuMeterWidget;
 
 use super::*;
 
+const AUDIO_CPU_INDICATOR_WIDTH: f32 = 62.0;
+const AUDIO_CPU_VALUE_WIDTH: f32 = 28.0;
+
 fn audio_stream_indicator(health: &AudioStreamHealth) -> Element<'_, Message> {
     let (color, description) = match health {
         AudioStreamHealth::Running => (th::success(), health.description().to_string()),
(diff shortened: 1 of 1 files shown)


alexanderwanyoike:

The fix itself is right: fixed 62px outer pill plus a fixed 28px right-aligned mono value box kills the digit-count jitter, and 999% (the clamp max) fits the value box in IBM Plex Mono.

1. **fix** `crates/vibez-ui/src/app/views_transport.rs:504` — The new test is tautological: it asserts the outer container width, which is unconditionally `Length::Fixed(AUDIO_CPU_INDICATOR_WIDTH)` regardless of load, so the loop over digit counts exercises nothing, and the inner `AUDIO_CPU_VALUE_WIDTH` box (the part that actually absorbs digit variation) is untested. Someone can delete the inner fixed-width box and the test still passes. Assert the value box's fixed width instead of (or in addition to) the outer.
2. **note** — The pill/tooltip chrome remains duplicated with `audio_stream_indicator` (and the perform views); this is the existing status-pill follow-up card from the v0.1.5 review, not new debt from this PR.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is. The extracted audio_cpu_value helper is now tested for the inner value-box width across digit counts (the part that actually absorbs variation), with a separate footprint test for the outer pill. Both findings closed.

PR #170, opened 2026-08-15
fix: keep Section record controls legible

alexanderwanyoike:
## Outcome

The shared Section Record/Capture strip now adapts before its controls or status text can be compressed.

## Root cause

The strip was permanently one 42 px-high row. Its fixed pickers/buttons consumed roughly 650 px before the flexible spacer, so opening the Browser in the reported 1400 px window left the target, phase, and Arrange destination to wrap into a squashed cluster.

## Fix

- `>= 1120 px`: retain the compact inline strip.
- `680–1119 px`: keep all controls on row one and move the complete status group to row two.
- `< 680 px`: split controls across two intentional rows and stack the status detail.

The width is the real Perform workspace after the Browser reservation, not the raw application window.

## Regression evidence

`record_bar_adapts_before_controls_or_status_can_be_squashed` failed against the old always-inline policy and now locks all three layout modes at 1400, 970, and 480 px.

Native QA reproduced the reported 1400 px window with Browser open: every control remains full-sized and all status text is visible on the second row.

## Verification

- `git diff --check`

### DIFF
--- a/crates/vibez-ui/src/app/views_perform_record.rs
+++ b/crates/vibez-ui/src/app/views_perform_record.rs
@@ -14,6 +14,9 @@ use crate::typography::{PERFORM_LABEL, PERFORM_TECH, PERFORM_TECH_STRONG};
 
 use super::*;
 
+const SECTION_RECORD_INLINE_MIN_WIDTH: f32 = 1_120.0;
+const SECTION_RECORD_STATUS_BELOW_MIN_WIDTH: f32 = 680.0;
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum SectionRecordBarLayout {
     Inline,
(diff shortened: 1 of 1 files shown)


alexanderwanyoike:
## mini-code-review — verdict: merge after fixing 1 and 2

1. **fix** `crates/vibez-ui/src/app/views_perform_record.rs:25` — The breakpoints `1_120.0` and `680.0` are inline magic numbers while the sibling `views_perform.rs` names every layout threshold as a top-of-file const (`MODE_TAB_MIN_WIDTH` etc.). Name them so they are visible to anyone scanning for layout constants.
2. **fix** — The 5-widget controls row (`row![record_button, count_in, mode, quantization, capture_button].spacing(7)`) is built verbatim in both the Inline and StatusBelow arms. Only one arm executes, so build it once before the match (widgets move, no Clone needed); StackedControls legitimately differs. As written, adding or reordering a control means editing three places.
3. **fix** (cheap ride-along) — `section_record_bar_layout` tests probe interior points (1400/970/480) but not the boundary contract; a `>=` vs `>` regression at either threshold passes. Assert at 1120/1119 and 680/679.
4. **note** — The three bar heights (42/64/104) are hand-estimated fixed pixels coupled to padding and font sizes defined elsewhere; a future tweak silently clips content. `Length::Shrink` is safe here (the workspace below is `Fill`) and would remove the coupling.

Verified clean: nothing depends on the old 42px height; the breakpoint basis correctly subtracts the browser dock; worst-case controls width fits the 680 breakpoint; in all three layouts overflow clips only status text, never controls.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is. All four findings closed, including the suggestion-level ones: breakpoints named and boundary-tested via the consts (exact >= contract pinned), controls row built once and shared across Inline/StatusBelow, and Length::Shrink replaces all three hand-estimated bar heights, removing the padding/font coupling entirely.

PR #172, opened 2026-08-15
Keep Browser WARP audition continuous

alexanderwanyoike:
## What changed
- play the truthful RAW source immediately while WARP awaits BPM detection or confirmation
- keep RAW audible while WARP is prepared, then crossfade through the existing Audition Bus
- distinguish `RAW · BPM NEEDED`, `RAW · WARPING`, and WARP playback in the footer
- make re-click and Play reuse decoded Local/Remote audio instead of silently doing nothing or decoding again
- rename `USE SOURCE` to the clearer `CONFIRM BPM`

WARP import remains blocked until source BPM is confirmed; this changes audition continuity, not the safety rule.

## Root cause
The WARP branch stopped the Audition Bus before checking whether a confirmed BPM existed. Low-confidence detections therefore left Browser silent until manual confirmation. Selecting an already-selected result was also a no-op.

## Verification
- red/green plan regression for unresolved WARP -> RAW Audition while import stays blocked
- regressions for decoded-source reuse and preparation that preserves the RAW voice/waveform
- native QA with `/home/alexander/Music/Samples/Synths/DD CHO 033.wav`: WARP + Loop progressed `RAW · BPM NEEDED` -> `RAW · WARPING` -> WARP after confirming its low-confidence 135.3 BPM suggestion
- test Local Root removed after QA
- full formatting, diff, strict workspace Clippy, and workspace test gate passed (541 UI tests)

### DIFF
--- a/crates/vibez-ui/src/app/engine_events.rs
+++ b/crates/vibez-ui/src/app/engine_events.rs
@@ -67,7 +67,10 @@ impl App {
                         self.state.browser.stop_audition_state();
                         if matches!(
                             self.state.status_text.as_str(),
-                            "RAW Audition playing" | "WARP Audition playing"
+                            "RAW Audition playing"
+                                | "WARP Audition playing"
+                                | "RAW Audition playing while WARP awaits source BPM"
+                                | "RAW Audition continues; confirm the edited BPM to hear WARP"
                         ) {
                             self.state.status_text = "Audition finished".into();
                         }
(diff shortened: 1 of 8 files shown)


alexanderwanyoike:
## mini-code-review — verdict: merge after fixing 1

The raw-voice-under-WARP design is right, and the scary paths check out: generation invalidation across re-selection, the single-slot waveform cache making a stale WARP result unable to land on a different source, the idempotent `ensure_raw_browser_audition`, and voice replacement riding the existing retired-slot mechanism.

1. **fix** `crates/vibez-ui/src/app/engine_events.rs:70` — The status literals ("RAW Audition playing while WARP awaits source BPM", "RAW Audition continues; confirm the edited BPM to hear WARP") are now duplicated between the sites that set them (`media.rs`, `update_media.rs`) and the string-match here that decides when to overwrite with "Audition finished". A wording tweak in one place silently breaks the transition. Extract shared constants (the match arm and the setters referencing the same `const`s).
2. **note** `crates/vibez-ui/src/app/update_media.rs:180` — "RAW Audition continues…" is set even when audition is disabled or no waveform is cached, i.e. when `ensure_raw_browser_audition` started nothing; the status then claims a voice that does not exist. Gate the wording on the ensure actually having a voice, or phrase it neutrally.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is. Status literals centralized as consts in app/mod.rs with an active_audition_status helper and a test enumerating the canonical statuses, so the Audition-finished transition can no longer silently break on a wording tweak. The truthfulness nit is fixed properly: ensure_raw_browser_audition now reports whether a voice is actually live and the status branches to a new confirm-BPM fallback when it is not.

PR #173, opened 2026-08-15
Fix deterministic Section split and delete

alexanderwanyoike:
## Outcome

Delete and Backspace now act on the fragment the producer visibly selected after splitting a Section clip, even when the previous edit was a time-range selection.

## Root cause

A normal clip click replaced `selected_clips` but left `time_selection_active` set. `SplitSelectedAtPlayhead` gives an active time range priority, so the same shortcut could perform either a direct clip split or a region split depending on hidden selection history. The resulting selection made the following Delete/Backspace appear intermittent.

## Change

- treat any clip click as an explicit return to clip-selection mode
- retire the stale time range, range track, and transient marquee before applying the clip selection
- cover the complete range → clip select → split → delete sequence with a regression test

## Verification

- native QA on a temporary copy of `test-perform-clips.vzp`: set edit cursor, drag a range, select the audio clip, split, Delete the left fragment, Backspace the remaining fragment
- source project was not modified
- `git diff --check`

### DIFF
--- a/crates/vibez-ui/src/domains/arrangement/clipboard.rs
+++ b/crates/vibez-ui/src/domains/arrangement/clipboard.rs
@@ -404,7 +404,7 @@ impl TimelineEditorState {
         let count = selected.len();
         self.selected_clips = selected;
         self.selected_track = Some(anchor_track);
-        self.time_selection_active = false;
+        self.clear_time_selection();
         ArrangementAction {
             status: Some(if count == 1 {
                 "Pasted clip".to_string()
--- a/crates/vibez-ui/src/domains/arrangement/mod.rs
+++ b/crates/vibez-ui/src/domains/arrangement/mod.rs
@@ -82,6 +82,12 @@ impl ProjectTracksState {
 }
 
 impl TimelineEditorState {
+    pub(super) fn clear_time_selection(&mut self) {
+        self.time_selection_active = false;
+        self.time_selection_track = None;
+        self.marquee = None;
+    }
+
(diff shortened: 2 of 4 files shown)


alexanderwanyoike:

The "deterministic" claim is honest: the nondeterminism source was hidden stale time-selection state, not collection order (the selected-clips HashSet iteration is order-invariant because each per-clip split/delete is independent).

1. **note** (follow-up PR per one-PR-per-issue) `crates/vibez-ui/src/domains/arrangement/mod.rs:770` — The same bug class survives via Ctrl+A: `SelectAllClips` replaces `selected_clips` but leaves `time_selection_active` set, so range-drag → Select All → `SplitSelectedAtPlayhead` still silently splits the stale region on `time_selection_track` instead of the visible selection (verified: the split guard at mod.rs:882 branches on `time_selection_active` first). Same one-line fix as this PR applies.
2. **note** `crates/vibez-ui/src/domains/arrangement/mod.rs:417` — The new inline triple-clear is the fourth ad-hoc variant of "clear time selection" (other sites clear different subsets of active/track/flag, and `op_delete_clips_in_region` leaves `time_selection_track` dangling). Each new call site re-deciding which fields to reset is exactly how this bug family keeps recurring; a `clear_time_selection()` helper on `TimelineEditorState` would pin the invariant.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is. clear_time_selection() now pins the invariant across all four previously ad-hoc sites, including the two that left time_selection_track dangling (paste and delete-region), and the existing test gained the track assertion. Merge this before #175, which uses the new helper.

PR #174, opened 2026-08-15
Polish live recording waveform rendering

alexanderwanyoike:
## Outcome

The part of a live Audio Input or MIDI Track Resample waveform that has already been drawn now stays visually fixed while new material grows onto the right edge.

## Root cause

The preview produces roughly 750 peak buckets per second, while the default timeline displays roughly 80 pixels per second. The renderer fitted the complete peak vector to the clip's current integer pixel width. Whenever the growing clip gained a pixel, historical columns were redistributed to different source buckets—the recorded audio was stable, but its picture jittered.

## Change

- carry the exact source-frame span represented by each live peak bucket into the timeline renderer
- resolve live waveform pixels from stable source-time ranges
- combine every bucket overlapping a pixel so downsampling keeps transient extrema
- preserve the existing fitted rendering for completed clips
- share the behavior across Audio Input and Track Output/Resample previews

## Regression evidence

- the focused regression failed with the same historical pixel jumping from bucket 100 to 94 as the take grew; it now remains mapped to source buckets 93–103
- peak compaction at the 4,096-bucket limit preserves the same displayed extrema
- completed clip sampling retains the previous width-fitted behavior

## Verification

- `git diff --check`
- `cargo test --workspace` (541 UI tests)

A native recheck was attempted against a temporary project copy, but the desktop session locked before the recording interaction. No source project was modified; visual dogfood remains the review step for this polish PR.


alexanderwanyoike:
## mini-code-review — verdict: merge after fixing 1

Per-frame costs verified sound: no O(captured_frames) work in the draw path (bounded by the 4096-peak cap, each peak visited at most ~twice), no allocations in the per-pixel loop, no audio-thread involvement, and the fallback path is behavior-identical to the old sampling.

1. **fix** `crates/vibez-ui/src/domains/audio_recording.rs:56` + `widgets/timeline/clips_draw.rs:69` — The live waveform leaves a growing blank strip at the clip's leading edge, exactly where the user watches while recording. `extend` only publishes a peak every `frames_per_peak` frames, so up to `frames_per_peak - 1` frames sit in the pending accumulator and are absent from `peaks`, while `duration` still includes them; pixels mapping past `peaks.len()` return None and the draw loop breaks. With `frames_per_peak` doubling at each compaction (64 → ~65k after an hour at 44.1kHz), the gap grows from invisible to ~1.5s (~150px at typical zoom). New in this PR: the old width-fitted sampling stretched available peaks across the full clip width. Fix is small: fold the pending partial peak into the `AudioRecordingPreview` snapshot as a synthetic final peak.
2. **note** — Two sampling models now coexist behind `Option<peak_span_frames>` (exact frame-span for live, width-fitted for persisted clips); unification would not be exactly behavior-preserving due to per-peak rounding in `compute_clip_peaks`, so this is documented maintenance cost, not a defect.

alexanderwanyoike:
## mini-code-review follow-up — verdict: merge as-is. The trailing-gap fix is elegant: the pending bucket is published eagerly and updated in place, compaction provably never sees a partial peak (it fires in the completion branch before the next bucket opens), and the new test pins the compaction+partial interaction on a long-take boundary. Verified the partial-peak min/max arithmetic by hand.

PR #178, opened 2026-08-16
fix(browser): make warp audition automatic

alexanderwanyoike:
## Summary

- replace Browser BPM detection and confirmation with deterministic complete-loop grid fitting across 1, 2, 4, 8, and 16 bars
- make WARP automatically target Project BPM, start on the next bar while transport runs, and loop; RAW stays immediate and one-shot
- use the exact same grid-fit treatment for audition, drag previews, local imports, and remote imports
- simplify the Audition footer to on/off, RAW/WARP, transport/waveform, and gain

## Root cause

Browser WARP exposed implementation uncertainty as a required producer workflow. Onset-derived BPM confidence gated both playback and import, while separate sync and loop controls allowed contradictory states. A valid loop could therefore be selected but remain silent until USE SOURCE or BPM confirmation was pressed.

## Verification

- cargo test --workspace
- native application smoke test at the maximized Browser layout

### DIFF
--- a/crates/vibez-engine/src/commands.rs
+++ b/crates/vibez-engine/src/commands.rs
@@ -13,12 +13,11 @@ use vibez_instruments::Instrument;
 
 use crate::playback_source::PreparedSectionPlaybackSource;
 
-/// Quantization policy for the next Browser Audition start.
+/// Automatic start policy selected by the Browser audition treatment.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum AuditionSync {
-    Off,
-    Beat,
-    Bar,
+pub enum AuditionStart {
+    Immediate,
+    NextBar,
 }
 
 /// Commands sent from the UI thread to the audio engine (via rtrb).
@@ -309,18 +308,16 @@ pub enum EngineCommand {
 
     // -- Dedicated Audition Bus --
     /// Request Browser Audition playback after project master processing.
-    /// Beat/Bar quantization applies only while transport is already running.
+    /// Next-bar scheduling applies only while transport is already running.
     StartAudition {
         audio: Arc<DecodedAudio>,
-        sync: AuditionSync,
+        start: AuditionStart,
         looped: bool,
     },
     /// Stop the Audition Bus with a short click-safe fade.
     StopAudition,
     /// Set Audition Bus gain independently from project mixer state.
     SetAuditionGain(f32),
-    /// Change looping for the active or queued Browser Audition.
-    SetAuditionLoop(bool),
 
     // -- External MIDI input --
     /// Route a live note-on from an external MIDI source (hardware
--- a/crates/vibez-engine/src/engine.rs
+++ b/crates/vibez-engine/src/engine.rs
@@ -10,7 +10,7 @@ use vibez_core::time::TempoMap;
 use vibez_dsp::factory::create_effect;
 use vibez_instruments::create_instrument;
 
-use crate::commands::{AuditionSync, EngineCommand};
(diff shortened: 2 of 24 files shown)


alexanderwanyoike:
Transport-sync follow-up pushed in 66d2a72.

Root cause confirmed by a failing engine repro: the loop-boundary crossfade skipped 220 source frames at 44.1 kHz on every wrap, so WARP previews ran about 5 ms short per loop and progressively drifted ahead. Absolute transport-position queues could also be stranded when an Arrange loop or Perform clock wrapped.

The fix now:
- preserves the exact warped loop length at every wrap
- schedules by elapsed frames so Arrange/Perform wraps cannot lose the launch
- rephases an already-running WARP preview when transport starts
- uses the active Arrange or Perform clock
- defaults normal Browser audition to synchronized WARP; RAW remains explicit

Regression coverage reproduces all three timing cases. Full workspace tests and warning-denied Clippy pass.

alexanderwanyoike:
Double-start follow-up pushed in baee214. Confirmed root cause: WARP preparation emitted an immediate provisional RAW audition, then emitted the real synchronized WARP launch at the bar. The provisional voice is removed; preparation is now silent and each selection produces one audible WARP start. Workspace tests and warning-denied Clippy pass.

alexanderwanyoike:

The engine work here is excellent and fixes a latent bug in passing: converting queued auditions from absolute target positions to transport-relative countdowns means an arrangement loop wrap can no longer strand a queued audition (new test proves it). Re-phase-on-transport-start is verified with marker-frame tests; the loop-wrap change correctly trades the old crossfade for grid-exactness (a synced loop must consume exactly one source length per wrap or it drifts, and the comment says so); `audition_playing` becoming event-driven instead of optimistic is strictly more truthful; and removing the persisted `audition_loop` field is serde-safe for old ui.json files.

### Design

1. **fix** `crates/vibez-ui/src/warp.rs` (`fit_loop_to_project`) + `crates/vibez-ui/src/state/mod.rs` (AuditionMode default) — Non-loop audio always gets warped, and WARP is now the default mode. The fit deliberately returns a result even when NO candidate lands in the plausible 60-200 BPM band, so a 0.4 s one-shot "fits" 1 bar at ~600 BPM, gets stretched ~5x, and loops on the bar grid; imports inherit this because `resolve_for_audio` only errors on empty audio. With Warp as the new default, that is the out-of-box preview for every one-shot library. The `has_plausible` flag already exists: when it is false, return None so the audition plan and the import treatment fall back to RAW for exactly the audio that is not a loop. One conditional; the determinism doctrine survives where it is true (real loops).

### Nits

2. **note** — Two BPM models now coexist: browser audition/import use the grid fit while OS drag-drop auto-warp still uses the detector + confidence threshold, so the same WAV can get different tempi depending on entry path, and the Settings "Warping" section now governs only the drop path. Unification is a follow-up card.
3. **note** — A seek during a queued countdown keeps counting in block time, so the start lands off-grid relative to the new position until the next transport start re-syncs it. The old absolute-position code had the complementary (worse) flaw; acceptable, just known.
4. **note** — RAW loop preview is removed entirely (looping is now WARP-implied) and the default mode flip changes browsing muscle memory; worth a line in the release notes.

alexanderwanyoike:
Review fix pushed in 809cdd2.\n\nAudio with no plausible 60–200 BPM loop interpretation now falls back to RAW for both audition and import instead of being stretched into an arbitrary bar. Real loops still receive the deterministic grid fit, and drag previews now follow the same fallback treatment.\n\nAdded one-shot regression coverage. Full workspace tests and warning-denied Clippy pass.

PR #179, opened 2026-08-16
style(browser): slim tree and results scrollbars

alexanderwanyoike:
## Summary

- use a shared slim vertical scrollbar configuration in the Browser
- reduce Browser tracks to 5 px and handles to 3 px
- apply the styling consistently to Places, local results, and remote results

## Tests

- cargo test --workspace

## Dependency

Stacked on #178.

### DIFF
--- a/crates/vibez-ui/src/app/views_browser.rs
+++ b/crates/vibez-ui/src/app/views_browser.rs
@@ -550,9 +550,7 @@ impl App {
                     SampleBrowserMode::Local,
                 ))
                 .height(Length::Fill)
-                .direction(scrollable::Direction::Vertical(
-                    scrollable::Scrollbar::default()
-                ))
+                .direction(scrollable::Direction::Vertical(browser_vertical_scrollbar()))
             ]
             .spacing(0)
             .padding(8)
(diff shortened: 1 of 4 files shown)


alexanderwanyoike:

1. **note** `crates/vibez-ui/src/app/views_browser_style.rs:19` — `const _: () = assert!(BROWSER_SCROLLBAR_TRACK_WIDTH <= 5.0)` guards a constant against the very literal it is defined as three lines above; a future widening hits a compile error with no stated rationale. Delete it or comment why 5 is the ceiling. (The HANDLE < TRACK assert is good and should stay.)
2. **note** — This is now the fifth ad-hoc `Scrollbar::new().width(w).scroller_width(s)` construction in the app (views_devices.rs x2, views_overlays.rs, views_perform_sections.rs); widths differ per site so they may be deliberate, but a shared theme-level constructor would keep scrollbars consistent. Follow-up candidate.

Verified: all three browser-view scrollables converted, none missed; the helper is width-only so scrollbar colors/behavior are untouched.

alexanderwanyoike:
Review cleanup pushed and the branch rebased onto the latest #178.\n\nRemoved the redundant compile-time assertion against the literal 5 px track width; the useful handle-smaller-than-track invariant remains. Full workspace tests and warning-denied Clippy pass.

PR #180, opened 2026-08-16
feat(browser): show audition waveform playhead

alexanderwanyoike:
## Summary

- report the active audition source position from the audio engine
- render a live playhead over the Browser waveform while auditioning
- reset and hide the playhead when an audition stops or changes

## Tests

- cargo test --workspace

## Dependency

Stacked on #178.

### DIFF
--- a/crates/vibez-engine/src/engine_audition.rs
+++ b/crates/vibez-engine/src/engine_audition.rs
@@ -183,6 +183,9 @@ impl AuditionBus {
         }) {
             self.active = None;
         }
+        if let Some(active) = self.active.as_ref() {
+            let _ = event_tx.push(EngineEvent::AuditionPosition(active.position));
+        }
         if had_voice && self.active.is_none() && !self.has_outgoing() && self.queued.is_none() {
             let _ = event_tx.push(EngineEvent::AuditionStopped);
         }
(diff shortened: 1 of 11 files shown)


alexanderwanyoike:

1. **fix** `crates/vibez-ui/src/widgets/browser_waveform.rs:95` — The playhead block is a near-verbatim copy of `widgets/waveform.rs:134` (same uncached second-Frame pattern, same 1.5-width `theme::playhead()` line), and it replicates that widget's quirk of allocating an empty Frame + geometry on every draw when no playhead is visible. Extract a shared `playhead_geometry(renderer, bounds, x: Option<f32>)` helper in `widgets/` used by both, which also fixes the empty-frame waste in one move.

Verified sound (all against actual code): the playhead is engine-truth, not UI-estimated — `EngineEvent::AuditionPosition` is pushed from the actual voice frame counter once per block (lock-free, matching the TrackMeter event precedent, drained at 60fps); the fraction divides by the buffer the engine actually plays, and since the WARP render is a uniform stretch it maps 1:1 onto the displayed raw waveform; loop wraps snap back correctly because the voice position resets mid-block; stale positions cannot resurrect a hidden playhead (FIFO ordering + `audition_playing` gate); and the expensive waveform summarization stays inside the canvas cache, so only the one-line playhead frame redraws per frame.

alexanderwanyoike:
Review fix pushed and the branch rebased onto the latest #178.\n\nExtracted one shared waveform playhead geometry helper and moved both the arrangement waveform and Browser waveform onto it. The helper returns no geometry when the playhead is hidden, so the Browser no longer allocates an empty overlay frame on every draw.\n\nFull workspace tests and warning-denied Clippy pass.

PR #184, opened 2026-08-16
Add Windows ASIO backend selection

alexanderwanyoike:
## What changed

- add a persisted audio backend choice to Soundcard Settings
- keep ALSA on Linux and CoreAudio on macOS while exposing WASAPI and ASIO on Windows
- remember System and ASIO devices separately, including migration from existing settings
- open ASIO input and output through the same duplex driver object
- release the exclusive ASIO driver before reconfiguration and restore the previous output after a rejected switch
- surface an input reopen failure after either a successful switch or an output rollback
- present one Audio Device choice for ASIO because one driver owns both directions
- hide Driver Type when the platform only has one backend
- replace the non-functional Hardware Setup button with clear ASIO routing guidance

## Dependency and SDK decision

CPAL stays on 0.16 and midir stays on 0.10. CPAL 0.16 already supports ASIO, so this PR does not need to replace the ALSA, CoreAudio, WASAPI and MIDI host layers at the same time. A CPAL upgrade needs a separate hardware regression pass.

Windows builds use the current Steinberg ASIO SDK under its GPLv3 licensing option. CI and release builds download it from Steinberg through one setup action, verify SHA-256 `d5ebf0c20dd2c5f43771fd0c1418f4b361bf52434ee670097cfa6b3a335e2eca`, cache that exact copy, and set `CPAL_ASIO_DIR` before Cargo runs. This prevents the mutable download inside `asio-sys` from deciding the release input.

The lockfile remains on the versions from `dev` apart from the ASIO feature dependencies and the serde dependencies used for audio backend settings.

## Why

A Windows producer using ASIO4ALL and a Yamaha or Steinberg UR12 could not see either ASIO driver in Vibez. The app only enumerated the default Windows host, WASAPI.

## Verification

- `cargo test --workspace --offline`
- `cargo check -p vibez-ui --offline`
- `cargo clippy --workspace --all-targets --exclude vibez-plugin-host --offline -- -D warnings`
- `cargo clippy -p vibez-plugin-host --offline -- -D warnings`
- `git diff --check`
- Windows CI asserts that the ASIO host is compiled and builds through the pinned SDK setup
- ASIO4ALL output selection was dogfooded on Windows; real UR12 and ASIO recording still need hardware confirmation

### DIFF
--- a/.github/actions/setup-asio-sdk/action.yml
+++ b/.github/actions/setup-asio-sdk/action.yml
@@ -8,7 +8,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: .asio-sdk
-        key: asio-sdk-d5ebf0c20dd2c5f43771fd0c1418f4b361bf52434ee670097cfa6b3a335e2eca
+        key: asio-sdk-${{ hashFiles('scripts/setup_asio_sdk.ps1') }}
     - name: Download and verify ASIO SDK
       shell: pwsh
       run: ./scripts/setup_asio_sdk.ps1
--- a/.github/actions/setup-asio-sdk/action.yml
+++ b/.github/actions/setup-asio-sdk/action.yml
@@ -1,0 +1,14 @@
+name: Set up pinned ASIO SDK
+description: Restore or verify the GPLv3-compatible Steinberg ASIO SDK used by CPAL
+
+runs:
+  using: composite
+  steps:
+    - name: Restore ASIO SDK cache
+      uses: actions/cache@v4
+      with:
+        path: .asio-sdk
+        key: asio-sdk-d5ebf0c20dd2c5f43771fd0c1418f4b361bf52434ee670097cfa6b3a335e2eca
+    - name: Download and verify ASIO SDK
+      shell: pwsh
+      run: ./scripts/setup_asio_sdk.ps1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
+      - name: Configure Windows ASIO build
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: '"LIBCLANG_PATH=C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_ENV -Append'
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
@@ -35,6 +39,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
+      - name: Configure Windows ASIO build
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: '"LIBCLANG_PATH=C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_ENV -Append'
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-      - name: Configure Windows ASIO build
+      - uses: ./.github/actions/setup-asio-sdk
         if: runner.os == 'Windows'
-        shell: pwsh
-        run: '"LIBCLANG_PATH=C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_ENV -Append'
+      - uses: Swatinem/rust-cache@v2
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
@@ -38,11 +36,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-      - name: Configure Windows ASIO build
+      - uses: ./.github/actions/setup-asio-sdk
         if: runner.os == 'Windows'
-        shell: pwsh
-        run: '"LIBCLANG_PATH=C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_ENV -Append'
+      - uses: Swatinem/rust-cache@v2
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,11 +49,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-      - name: Configure Windows ASIO build
+      - uses: ./.github/actions/setup-asio-sdk
         if: runner.os == 'Windows'
-        shell: pwsh
-        run: '"LIBCLANG_PATH=C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_ENV -Append'
+      - uses: Swatinem/rust-cache@v2
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,6 +50,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
+      - name: Configure Windows ASIO build
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: '"LIBCLANG_PATH=C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_ENV -Append'
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+/.asio-sdk
 *.swp
 *.swo
 *~
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,9 +65,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alsa"
-version = "0.11.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "812947049edcd670a82cd5c73c3661d2e58468577ba8489de58e1a73c04cbd5d"
+checksum = "ed7572b7ba83a31e20d1b48970ee402d2e3e0537dcfe0a3ff4d6eb7508617d43"
 dependencies = [
  "alsa-sys",
  "bitflags 2.11.0",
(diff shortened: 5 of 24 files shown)


alexanderwanyoike:
Windows ASIO test build is ready: [download the `vibez-windows-x86_64` artifact](https://github.com/alexanderwanyoike/vibez/actions/runs/31966744004/artifacts/9268801392). The artifact contains both the setup executable and portable ZIP built from commit `a4eb80d`.

alexanderwanyoike:
Windows dogfood fixes are now in `863bdfb`:

- ASIO-to-ASIO switches retry the requested driver with its own defaults.
- If both attempts fail, Vibez reopens the exact output configuration that was running before the switch.
- Settings now report the actual live backend/device after rejection.
- ASIO routing guidance points users to the driver control panel for speakers/headphones.
- CPAL is upgraded to 0.17.3 for the ASIO reset, shared-driver and non-conformant-driver fixes.

Local verification: full workspace tests pass, focused strict Clippy passes, and formatting/diff checks pass. Windows CI will exercise the ASIO build.

alexanderwanyoike:

Settle the ASIO SDK build/licensing question (finding 4) before this ships in a release, and fix findings 1 and 2, which are user-visible regressions in the new switching path.

## Bugs

1. **fix** — `crates/vibez-ui/src/app/audio_settings.rs:195` — Switching driver type discards and persists away the user's device selection. `handle_select_audio_backend` clears both `preferred_input_name` and `preferred_output_name`, then applies `apply_audio_output_configuration(None, ...)`; on success the handler persists `preferred_output_name = None` via `persist_ui_settings()`. A Windows user with a UR12 selected under WASAPI who toggles to ASIO to try it and toggles back lands on laptop speakers with the UR12 choice erased from `ui.json`. The round trip is lossy every time.

2. **fix** — `crates/vibez-ui/src/app/audio_settings.rs:448` — The input-reopen failure after a rejected output switch is swallowed: `let _ = self.sync_audio_input_runtime();`. The input stream was already torn down (line ~364) before the reconfigure attempt, so if the rollback restores the old output but the input cannot reopen on it, the status text says "<device> restored" while monitoring and arming are silently dead. The success path surfaces this error; the failure path drops it.

3. **fix** (PR body/scope, not code) — `Cargo.toml:33` — A cpal major bump (0.16 → 0.17.3) is bundled in unmentioned, dragging alsa 0.9 → 0.11, coreaudio-rs 0.13 → 0.14 (the objc2 rewrite), and midir 0.10 → 0.11. That is a rewrite of the Linux and macOS host layers landing under a "Windows ASIO" title; any Linux/macOS audio regression is invisible in release notes and hard to attribute when bisecting. At minimum the PR body should say so; ideally the bump is a separate, independently revertible PR.

## Design / DRY

4. **fix** (CI determinism) + **note** (licensing decision for the maintainer) — `crates/vibez-audio-io/Cargo.toml:20` — The ASIO SDK is downloaded from steinberg.net at build time and statically linked into a GPL binary. asio-sys 0.2.6's build script fetches `https://www.steinberg.net/asiosdk` into `%TEMP%\asio_sdk` whenever `CPAL_ASIO_DIR` is unset, and the workflows set only `LIBCLANG_PATH`. (a) Build: `Swatinem/rust-cache` does not cache `%TEMP%`, so every fresh runner refetches; a steinberg.net outage or URL change breaks all Windows CI and every release cut — pin/cache the SDK explicitly. (b) Licensing: the resulting `libasio.a` is compiled from Steinberg's proprietary SDK and linked into a binary distributed under GPL-3.0-or-later; this is the specific reason Audacity has never shipped ASIO-enabled binaries. The README trademark line covers attribution, not redistribution of a linked binary. Needs a maintainer decision before a release goes out.

5. **fix** — `crates/vibez-audio-io/src/audio_host.rs:285` — `impl Default for AudioHost` duplicates the System-backend mapping (`backend: System, host: cpal::default_host()`) instead of delegating to `Self::new(AudioBackend::System)`, and has no callers left in the workspace. Delete it or make it delegate.

6. **note** — `crates/vibez-ui/src/ui_settings.rs` — One device-name slot is shared by all backends; this is the root cause of finding 1 (a WASAPI device name is meaningless to the ASIO host, so the switch handler has no option but to wipe it). Per-backend remembered selections would remove both the clear and the startup fallback churn.

## Nits

7. **fix** — `crates/vibez-ui/src/app/views_settings_audio.rs:215` — Every Linux and macOS user now gets a "Driver Type" dropdown with exactly one unselectable entry plus a divider. Hide the row when `AudioBackend::available().len() == 1`.

8. **fix** — `crates/vibez-ui/src/app/update.rs:385` / `views_settings_audio.rs:233` — The "Hardware Setup" button only writes a status string telling the user to find the driver's control panel themselves. A button that looks like it opens something and doesn't will be reported as broken; a plain label sets expectations correctly.

9. **note** — `crates/vibez-ui/src/app/audio_recording.rs` is now 952 lines against the project's 1,000-line rule (this PR added ~33). Not a blocker, but the next change to that file has to split it first.

10. **note** — `Cargo.lock` — the regeneration flips wgpu/wgpu-core/wgpu-hal from parking_lot 0.12.5 to 0.11.2 (dev has them on 0.12.5) plus assorted windows-sys downgrades — unexplained resolution churn on packages unrelated to this PR. Prefer a minimal lock update (`cargo update -p cpal -p midir` from dev) so the diff only contains what the PR needs.

alexanderwanyoike:
Addressed the complete review in `6cd08c3`.

1. System and ASIO now have separate persisted input/output preferences. Switching back restores the previous device instead of erasing it.
2. A failed input reopen after output rollback is surfaced in the status, with a regression test for the combined message.
3. CPAL is back on 0.16 and midir on 0.10. The Linux and macOS host upgrades are no longer part of this PR.
4. Windows CI and release builds now use a checksum-pinned and cached Steinberg SDK through `CPAL_ASIO_DIR`. The current SDK has an explicit GPLv3 licensing option, which matches the project license. The README records that decision and the local setup command.
5. The unused `Default for AudioHost` implementation is removed.
6. Device persistence is modelled per backend, with old `ui.json` settings migrated into the saved backend.
7. Driver Type is hidden when only one backend exists.
8. Hardware Setup is plain routing guidance now, not a button that does nothing.
9. Recorded-take staging and its tests moved into `audio_take_finalization.rs`; `audio_recording.rs` is down from 952 to 727 lines.
10. The lockfile now matches `dev` apart from ASIO and audio-settings serialization dependencies.

Local verification is green for the full workspace tests, strict Clippy, formatting, and diff checks. The new three-platform CI run is in progress.

alexanderwanyoike:
Final verification is complete. Linux, macOS and Windows checks and tests all pass, including the Windows CPAL 0.16 ASIO build through the pinned SDK action. Clippy and formatting also pass.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge after fixing N1

All nine findings from the previous review are resolved (several more thoroughly than asked: per-backend device memory with legacy migration, the cpal 0.16 revert keeping asio-sys working, the checksum-verified SDK setup with a licence assertion, and the `audio_take_finalization.rs` split). CI is green at 6cd08c3. One new bug in the SDK pinning and a few cleanups.

1. Backend switch wiped device selection — **resolved** (`domains/audio_settings.rs:59` `AudioBackendPreferences`, `remember_current_backend`/`restore_backend_preferences`, persisted as `audio_device_preferences` with legacy migration at `ui_settings.rs:192`, round-trip + migration tests).
2. Swallowed input-reopen error — **resolved** (`app/audio_settings.rs:469` composes it via `rejected_output_status()`, tested; `persist_ui_settings()` also moved ahead of the input-sync early return so a successful output change is saved even when input can't reopen).
3. Unannounced cpal major bump — **resolved** (reverted to cpal 0.16 / midir 0.10; alsa 0.9.1 and coreaudio-rs 0.13.0 restored; asio-sys still present so ASIO works on 0.16; comment notes host upgrades belong in their own PR).
4. ASIO SDK download/licensing — **largely resolved** (`scripts/setup_asio_sdk.ps1` pins SHA256, asserts the archive's GPLv3 licence option, exports `CPAL_ASIO_DIR`; cached by `.github/actions/setup-asio-sdk`; README states the GPLv3 basis). Residual gaps are N1/N2 below.
5. Dead `impl Default for AudioHost` — **resolved** (deleted).
6. Shared device-name slot — **resolved** (same mechanism as 1).
7. Single-option Driver Type picker on non-Windows — **resolved** (row gated behind `AudioBackend::available().len() > 1`).
8. "Hardware Setup" button — **resolved** (button and message removed, static explanatory text instead).
9. `audio_recording.rs` near 1,000 lines — **resolved** (split to 727; take staging moved to `audio_take_finalization.rs`, 245 lines, with tests).
10. Lockfile parking_lot/windows-sys churn — **resolved** by the cpal revert.

## New findings

### Bugs

N1. **fix** — `scripts/setup_asio_sdk.ps1:6` — A SHA256 is pinned against the mutable alias URL `https://www.steinberg.net/asiosdk`, which always serves Steinberg's *current* SDK. When Steinberg publishes the next SDK the download stops matching and the script throws — asymmetrically: runners with a warm cache (key embeds the same checksum, `.github/actions/setup-asio-sdk/action.yml:13`) stay green while fresh runners and every local Windows developer go red, a partial CI outage with no code change to explain it. Pin a versioned artifact URL (Steinberg serves e.g. `https://download.steinberg.net/sdk_downloads/asiosdk_2.3.3_2019-06-14.zip`) so archive and checksum cannot drift apart.

### Design / DRY

N2. **fix** (doc or guard) — `docs/ARCHITECTURE.md:463` claims asio-sys "never performs its own mutable build-time download", but that holds only where `setup_asio_sdk.ps1` ran in the same shell. A Windows developer building from a fresh terminal, cmd.exe, or an IDE silently gets asio-sys's unpinned `%TEMP%` download — the exact behaviour this work removes — with no warning. Either fail fast when `CPAL_ASIO_DIR` is unset on Windows, or scope the doc claim to CI/release builds.

N3. **fix** — The SDK checksum is duplicated between `.github/actions/setup-asio-sdk/action.yml:13` (cache key) and `scripts/setup_asio_sdk.ps1:6`. Bumping the script alone leaves a stale cache key; `Test-PinnedSdk` then rejects the restored copy and re-downloads every run, so the cache silently stops working. Derive the key from the script (or one pinned file).

N4. **fix** — `crates/vibez-ui/src/app/audio_settings.rs:209` — `previous` is cloned *before* `remember_current_backend()` runs on the next line, so both failure rollbacks discard what that call just saved. In practice the preferences are already in sync there (every successful change calls `remember_current_backend()` and persists), which makes line 210 a no-op — so either it is dead and should go, or it is load-bearing and the rollback ordering is wrong. A maintainer cannot tell which from reading it.

### Nits

N5. **note** — `crates/vibez-audio-io/src/audio_host.rs:345` — `cpal_device_name()` is now just `device.name()`; it was introduced for the cpal 0.17 `description().name()` API and survived the revert. Inlining removes an indirection at six call sites that reads as if it does something.

N6. **fix** — `crates/vibez-ui/src/app/project_io.rs:279` still populates `preferred_audio_input`/`preferred_audio_output`, which are now `#[serde(skip_serializing)]` (`ui_settings.rs:49,55`). Both clones are computed and discarded at every save, and a reader would conclude the legacy fields are still written.

alexanderwanyoike:
Addressed every re-review finding in 9f99e48.

- The SDK now uses Steinberg’s immutable ASIO SDK 2.3.4 archive URL. I verified the downloaded archive still matches the pinned SHA256.
- The action cache key derives from the setup script, so URL and checksum changes invalidate it together.
- The architecture note now scopes the no-mutable-download guarantee to distributed builds and documents the local setup requirement.
- Backend rollback snapshots the preferences after capturing the current selection.
- The leftover CPAL device-name wrapper is removed.
- New saves leave the legacy skipped device fields empty and persist only per-backend preferences.

Full workspace tests, strict Clippy, formatting and diff checks pass.

alexanderwanyoike:
Final verification is green on all eight CI jobs, including the Windows ASIO build and full Windows test suite using the immutable SDK archive.

alexanderwanyoike:

All six round-2 findings are resolved in 9f99e48; no new findings. CI is green on all platforms, and the Windows job re-downloaded the SDK fresh (the new cache key invalidated the old cache), which live-verifies the versioned URL against the pinned SHA256.

- N1 (mutable alias URL vs pinned SHA) — **resolved**: `scripts/setup_asio_sdk.ps1:6` now pins the immutable `download.steinberg.net/sdk_downloads/ASIO-SDK_2.3.4_2025-10-15.zip`; the checksum is unchanged and the fresh CI download matched it.
- N2 (ARCHITECTURE.md overclaim) — **resolved**: the "never performs its own mutable build-time download" claim is now scoped to distributed builds, with the local-build requirement to run the script in-session stated alongside.
- N3 (checksum duplicated in cache key) — **resolved**: the action's cache key is now `hashFiles('scripts/setup_asio_sdk.ps1')`, so URL and checksum changes invalidate the cache together with a single source of truth.
- N4 (rollback clone ordered before `remember_current_backend()`) — **resolved**: the capture now runs first, so both failure rollbacks retain the just-remembered device choices; the comment states the constraint.
- N5 (vestigial `cpal_device_name`) — **resolved**: inlined to `device.name()` at all six call sites.
- N6 (legacy fields still populated) — **resolved**: `project_io.rs` writes `None` into the skip-serialized legacy fields with a comment marking them read-only migration inputs.

PR #185, opened 2026-08-16
Embed the Windows application icon

alexanderwanyoike:
## What changed

- embed `assets/icon/vibez.ico` into the Windows PE for the installed and portable executable
- add product metadata to the executable resource
- verify RT_GROUP_ICON resource ID 1 in Windows CI and release builds

## Why

The installer carried an icon, but `vibez.exe` itself contained no Windows resources. Explorer, Start menu and installed shortcuts therefore fell back to a generic executable icon.

## Verification

- `cargo check -p vibez --all-targets`
- `cargo test -p vibez --all-targets`
- `cargo clippy -p vibez --all-targets -- -D warnings`
- Windows CI builds the executable and calls `FindResourceW` for RT_GROUP_ICON ID 1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,6 @@ jobs:
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
       - run: cargo check --workspace
-      - name: Verify Windows executable resources
-        if: runner.os == 'Windows'
-        shell: pwsh
-        run: |
-          cargo build --package vibez --bin vibez
-          ./scripts/check_windows_resources.ps1 target/debug/vibez.exe
 
   test:
     name: Test
@@ -45,6 +39,12 @@ jobs:
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libasound2-dev libudev-dev libdbus-1-dev
       - run: cargo test --workspace
+      - name: Verify Windows executable resources
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          cargo build --package vibez --bin vibez
+          ./scripts/check_windows_resources.ps1 target/debug/vibez.exe
 
   clippy:
     name: Clippy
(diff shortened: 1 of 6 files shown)


alexanderwanyoike:

Clean. Verified: `assets/icon/vibez.ico` exists on dev, `build.rs` early-returns on non-Windows targets, keeping `winresource` as an unconditional build-dependency is correct (build-deps key on the host, so a cfg-gated dep would break cross-compilation), and release.yml builds without `--target`, so the `target/release/vibez.exe` path the check script uses is right. No bugs found; none of the nits block it.

## Nits

1. **note** — `.github/workflows/ci.yml:27` — the verification hangs off the *check* job, which otherwise only runs `cargo check`, so it adds a full debug link of the GUI binary to the slowest CI leg while the `test` job builds the same crate graph separately. Moving it to the `test` job or a dedicated Windows-only job avoids paying for the link twice.

2. **note** — `build.rs:11` sets `ProductName` to `"vibez"` and `scripts/check_windows_resources.ps1:52` asserts that literal. Renaming the product in one place fails CI with a message pointing at the script rather than the rename. Asserting non-empty, or reading the expected name from `cargo metadata`, removes the coupling.

3. **fix** (one-flag change) — `scripts/check_windows_resources.ps1:6` — `Resolve-Path $Executable` on a missing path emits a non-terminating error and leaves `$resolved` null, so a missing executable surfaces as a Win32 error out of `LoadLibraryExW` instead of "file not found". Add `-ErrorAction Stop`.

4. **note** — `Cargo.lock` carries an unrelated `windows-sys 0.59.0 → 0.60.2` bump for a transitive dep — harmless drift from a `cargo update`, but not part of embedding an icon.

alexanderwanyoike:
Addressed all review notes in ff36bbf.

1. Windows resource verification now runs in the Test job, where the binary is already linked.
2. The script verifies that ProductName is present without coupling itself to a duplicate literal.
3. Resolve-Path now uses ErrorAction Stop, so a missing executable reports the real failure.
4. The unrelated windows-sys lockfile drift is removed.

All eight CI jobs pass.

alexanderwanyoike:

All four nits from the previous review are resolved exactly as described, and the follow-up commit (ff36bbf) introduced nothing new. CI is green.

1. Resource check in the `check` job — **resolved** (moved to the `test` job, so the binary links once on the job that already builds it).
2. Hard-coded `ProductName` assertion — **resolved** (`scripts/check_windows_resources.ps1:53` now asserts non-empty).
3. `Resolve-Path` on a missing file — **resolved** (`-ErrorAction Stop` added).
4. Unrelated `Cargo.lock` drift — **resolved** (`windows-sys` back to 0.59.0; the lockfile change is now limited to adding `winresource`).

No new findings.

PR #186, opened 2026-08-16
Render Browser tree icons consistently on Windows

alexanderwanyoike:
## What changed

- replace raw Unicode circles, chevrons, dots and refresh markers in the Browser Places tree
- render each marker through the bundled Lucide icon font
- add the missing Lucide codepoint constants

## Why

Those markers relied on platform font fallback. Windows could not resolve them from Vibez's text fonts and displayed box glyphs. The Browser now renders the same icons on Linux, macOS and Windows.

## Verification

- no platform-dependent tree marker glyphs remain in Browser views
- all 556 `vibez-ui` library tests pass
- strict `vibez-ui` Clippy passes
- formatting and diff checks pass


alexanderwanyoike:
# mini-code-review — verdict: merge after collapsing the duplicated icon blocks (finding 3); the rest are follow-ups

All three new codepoints verified against the bundled `assets/fonts/Lucide.ttf` cmap: `E06F` → `chevron-right`, `E44F` → `dot`, `E145` → `refresh-cw`. The diff is mechanical and clippy-clean, and it is a strict improvement on the reported Windows bug. No bugs found.

## Design / DRY

1. **note** (follow-up card) — `crates/vibez-ui/src/app/mod.rs:122` — The root cause is unaddressed, so the same class of bug remains. The application builder never calls `.default_font(...)`; it only registers fonts, so every bare `text(...)` resolves through `Font::DEFAULT` → system sans-serif via fontdb — exactly the platform-fallback path this PR is fixing. Still on that lottery: `—` at `views_browser.rs:592`/`:596` and `views_browser_remote.rs:336`, `·` at `views_perform_pads.rs:148`, `›` at `views_file_menu.rs:74` and `views_perform_record.rs:359`, `…` in `views_settings_*`. Setting `.default_font(PLEX_SANS_CONDENSED_MEDIUM)` fixes the whole class at once (the bundled IBM Plex TTFs cover U+2014/U+00B7/U+203A/U+2026, but not the replaced U+25BE/25B8/25CF/25CB — so the icon swap was the right call for the chevrons and circles specifically).

2. **note** (follow-up card) — `crates/vibez-ui/src/app/views_browser_places.rs:23` — `CIRCLE` vs `CIRCLE_DOT` at `size(9)` loses most of the active/inactive contrast: Lucide's stroked ring is ~0.75 px at that size and the distinguishing dot ~1.5 px, where the old `●`/`○` pair was solid-disc versus hollow. Color still carries the state, but the shape signal is nearly gone and the ring will alias on non-HiDPI. Same applies to the chevrons at `size(10)`; Lucide's 24 px stroked geometry generally wants 12–14 to read at the weight the replaced glyphs had.

3. **fix** — `crates/vibez-ui/src/app/views_browser_places.rs:88`, `:105`, `:157`, `:198`, `:319`, `:334`, `:394`, `:409` — The toggle and leaf-marker blocks are copy-pasted 5 and 3 times: each chevron site is an identical seven-line `icons::icon(if expanded { CHEVRON_DOWN } else { CHEVRON_RIGHT }).size(10).color(th::text_muted())`, each leaf a `container(icons::icon(icons::DOT)...)`. The duplication predates the PR but the PR roughly doubles each instance's size, so any future size/color tweak now has eight edit sites. Two private helpers in this file collapse it.

## Nits

4. **note** — `crates/vibez-ui/src/app/views_browser_places.rs:135` — `state_marker` keeps a raw `text("!")` while its two siblings became icons. It's ASCII so it renders fine, but the branch that most wants a recognizable glyph (`STALE`/`WATCH ERR`/`WARN`) is now the only non-icon; Lucide ships `triangle-alert`/`circle-alert`, and a `match` returning a codepoint would restore the original compact shape.

alexanderwanyoike:
Addressed the complete review in 1246c89.

1. IBM Plex is now the application default font, removing the system-font fallback for ordinary punctuation.
2. Browser tree icons are 12 px so their stroke and state differences remain readable.
3. Repeated toggle, leaf and state-marker blocks are collapsed into private helpers.
4. Warning states use the Lucide triangle-alert icon instead of a raw exclamation mark.

All eight CI jobs pass.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge as-is *if* the default-font switch is intentional and someone has eyeballed the app

The icon work is complete and correct: helpers extracted, sizes raised to 12, `triangle-alert` (verified in the bundled Lucide TTF) for the warning marker. Suites pass (556 vibez-ui tests), clippy clean. The one open question is scope, not correctness.

1. No `default_font`, glyph-fallback class remains — **resolved with a caveat** (`app/mod.rs:152` sets `.default_font(PERFORM_DISPLAY)`, fixing the root cause for every bare `text()`; see N1 on the choice of face).
2. Icon contrast at size 9/10 — **partially resolved** (named constants at 12.0; ring stroke ~1 px and dot ~2 px, above the sub-pixel threshold; state contrast is still ring-vs-ring-plus-dot rather than filled-vs-hollow, but color carries it — a reasonable place to stop).
3. Chevron/leaf duplication — **resolved** (four free functions — `place_mode_marker`, `tree_toggle`, `tree_leaf_marker`, `catalog_state_marker` — replace all five chevron sites and three leaf sites; file net 19 lines shorter).
4. `text("!")` among icon siblings — **resolved** (`TRIANGLE_ALERT` + a `match` returning `(codepoint, color)`).

## New findings

### Design

N1. **note** (maintainer decision) — `crates/vibez-ui/src/app/mod.rs:152` — `.default_font(PERFORM_DISPLAY)` is a whole-app typography change shipped inside a Browser tree-icon PR. It is the right mechanism, but `PERFORM_DISPLAY` is IBM Plex Sans **Condensed** Medium — a role `typography.rs` documents as scoped to the Perform surface. Promoting it to the global default makes the 13 explicit `.font(PERFORM_DISPLAY)` sites no-ops and switches roughly 500 previously system-font `text()` call sites to a condensed face in one commit; condensed metrics are materially narrower, so every `Length::Fixed` text cell changes its fill ratio (e.g. the 36/50/48 px BPM/length/source columns at `views_browser.rs:420`). Either split into its own PR with a screenshot, or bundle a non-condensed Plex Sans as the default text role and keep Condensed scoped to Perform.

### Nits

N2. **note** — `views_browser_places.rs:31`/`:45` — the shared helpers use `padding([3, 2])`/`[3, 3]` where the root-level sites they replaced used `[4, 2]`/`[4, 3]`. Verified harmless for layout (row is center-aligned and `root_button` still carries `[4, 2]`), but the root toggle's click target is 2 px shorter, unremarked in the commit.

N3. **note** — `views_browser_places.rs:16` — `PLACE_STATE_ICON_SIZE` and `TREE_MARKER_SIZE` are both `12.0` and used in the same tree; two names for one value invites drift until there's an actual reason to split them.

alexanderwanyoike:
Addressed the re-review notes in b75f285.

- Removed the whole-app PERFORM_DISPLAY default. This PR now fixes the reported Windows glyphs without switching roughly 500 unrelated text sites to a condensed face.
- Restored the original tree-toggle and leaf-marker padding, preserving the prior hit targets.
- Consolidated the identical 12 px tree icon constants.

All 556 vibez-ui tests, strict Clippy, formatting and diff checks pass.

alexanderwanyoike:
Final verification is green on all eight Linux, macOS and Windows CI jobs.

alexanderwanyoike:

All three round-2 findings are resolved in b75f285; no new findings. CI green.

- N1 (app-wide `.default_font(PERFORM_DISPLAY)` condensed-font switch) — **resolved**: the global default was dropped entirely, keeping the PR scoped to the Browser tree icons. This is safe because every marker this PR touches now goes through `icons::icon()`, which sets the bundled Lucide font explicitly — the icons no longer depend on fallback, so the app's text typography stays exactly as on dev.
- N2 (helper paddings shrank the root toggle's click target) — **resolved**: `tree_toggle`/`tree_leaf_marker` restored to `[4, 2]`/`[4, 3]`.
- N3 (two constants for one value) — **resolved**: merged into a single `TREE_ICON_SIZE`.

PR #187, opened 2026-08-16
Fix Browser warp tempo analysis

alexanderwanyoike:
## What changed

- derive a loop tempo from its source boundary, filename BPM hint, and existing onset analysis
- remove Project BPM from the source-tempo fitting API
- analyse local previews off the UI thread and reuse revision-matched remote derived metadata for both audition and import
- cache one intrinsic loop fit for Audition, drag previews, BPM display, and imports
- ignore filename tempo hints that disagree with the complete source length and fall back to audio evidence
- align onset octave folding and loop fitting at a 200 BPM upper bound

## Why

The old fit selected whichever bar count required the least stretching toward the current Project tempo. A real two-bar 124 BPM UNDGRND loop was therefore interpreted as one bar at 62 BPM in an 80 BPM Project. Source identity must not change with Project tempo.

The onset fold upper bound moves from 180 to 200 BPM. Existing cached remote analysis at roughly 181 to 200 BPM can retain its previous half-time value until that provider revision is rematerialized. Audition and import intentionally share that revision-matched cached value, so a sample never changes tempo between preview and placement.

## Verification

- cargo test --workspace
- cargo clippy --workspace --all-targets --exclude vibez-plugin-host -- -D warnings
- cargo fmt --all -- --check
- regression covers the reproduced 170,710-frame, 44.1 kHz, two-bar 124 BPM sample shape at both 80 and 160 BPM Project tempos
- regression covers a cached 95 BPM remote audition importing at the same tempo even when fresh analysis detects 190 BPM

### DIFF
--- a/crates/vibez-core/src/onset.rs
+++ b/crates/vibez-core/src/onset.rs
@@ -168,6 +168,8 @@ const BPM_MIN: f64 = 40.0;
 const BPM_MAX: f64 = 280.0;
 // Target octave for folded output (dance-music conventional range).
 const BPM_FOLD_LO: f64 = 60.0;
+// Keep aligned with vibez-ui's PLAUSIBLE_LOOP_BPM_MAX so Browser fitting does
+// not accept a source tempo which onset analysis has already folded in half.
 const BPM_FOLD_HI: f64 = 200.0;
 const BPM_FOLD_PREF: f64 = 120.0;
 // Working sample rate for the autocorrelation. 500 Hz gives a lag
--- a/crates/vibez-core/src/onset.rs
+++ b/crates/vibez-core/src/onset.rs
@@ -11,7 +11,7 @@
 //!   Drives slice-and-snap audio quantize.
 //! - `detect_bpm` returns an `Option<BpmEstimate>`. Autocorrelates the
 //!   onset flux (so it works on sustained melodic material, not just
-//!   percussion), octave-folds into [60, 180] BPM using Parncutt's
+//!   percussion), octave-folds into [60, 200] BPM using Parncutt's
 //!   preference curve, and gates on confidence so sparse or silent
 //!   clips return `None` rather than guessing.
 
@@ -168,7 +168,7 @@ const BPM_MIN: f64 = 40.0;
 const BPM_MAX: f64 = 280.0;
 // Target octave for folded output (dance-music conventional range).
 const BPM_FOLD_LO: f64 = 60.0;
-const BPM_FOLD_HI: f64 = 180.0;
+const BPM_FOLD_HI: f64 = 200.0;
 const BPM_FOLD_PREF: f64 = 120.0;
 // Working sample rate for the autocorrelation. 500 Hz gives a lag
 // precision of ~0.5 BPM at 120 BPM, further refined by parabolic
@@ -186,7 +186,7 @@ const MIN_SECONDS_FOR_BPM_F64: f64 = 1.0;
 
 /// Detect the tempo of `audio`. Returns `None` when the audio is too
 /// short, too sparse, or the autocorrelation is not strong enough to
-/// commit to a tempo. Output BPM is always in `[60, 180]` after
+/// commit to a tempo. Output BPM is always in `[60, 200]` after
 /// Parncutt octave folding.
 pub fn detect_bpm(audio: &DecodedAudio, sample_rate: u32) -> Option<BpmEstimate> {
     let frames = audio.num_frames();
@@ -616,6 +616,17 @@ mod tests {
         assert!(
             (est.bpm - 174.0).abs() < 1.5,
             "expected ~174 BPM, got {}",
+            est.bpm
+        );
+    }
+
+    #[test]
+    fn detects_190_bpm_without_folding_to_half_time() {
+        let audio = synthetic_kick_track(44_100, 190.0, 8.0);
+        let est = detect_bpm(&audio, 44_100).expect("expected detection");
+        assert!(
+            (est.bpm - 190.0).abs() < 2.0,
+            "expected ~190 BPM, got {}",
             est.bpm
         );
     }
--- a/crates/vibez-ui/src/app/async_helpers.rs
+++ b/crates/vibez-ui/src/app/async_helpers.rs
@@ -36,8 +36,24 @@ pub(super) async fn analyse_browser_audio_async(
     audio: vibez_core::audio_buffer::DecodedAudio,
     source_name: String,
 ) -> Result<crate::message::AnalysedBrowserAudio, String> {
+    analyse_browser_audio_with_cached_metadata_async(audio, source_name, None).await
+}
+
+async fn analyse_browser_audio_with_cached_metadata_async(
+    audio: vibez_core::audio_buffer::DecodedAudio,
+    source_name: String,
+    cached_metadata: Option<vibez_dropbox::DerivedMetadata>,
+) -> Result<crate::message::AnalysedBrowserAudio, String> {
     tokio::task::spawn_blocking(move || {
-        let loop_fit = crate::warp::analyse_loop_tempo(&audio, &source_name);
+        let loop_fit = match cached_metadata {
+            Some(metadata) => crate::warp::fit_loop_tempo(
+                audio.num_frames(),
+                audio.sample_rate as f64,
+                &source_name,
+                metadata.bpm,
+            ),
+            None => crate::warp::analyse_loop_tempo(&audio, &source_name),
+        };
         crate::message::AnalysedBrowserAudio {
             audio: Arc::new(audio),
             loop_fit,
@@ -45,6 +61,14 @@ pub(super) async fn analyse_browser_audio_async(
     })
     .await
     .map_err(|error| format!("Browser analysis task failed: {error}"))
+}
+
+pub(super) async fn decode_and_analyse_async(
+    path: PathBuf,
+    source_name: String,
+) -> Result<crate::message::AnalysedBrowserAudio, String> {
+    let audio = decode_file_async(path).await?;
+    analyse_browser_audio_async(audio, source_name).await
 }
 
 pub(super) async fn decode_analyse_and_stage_local_async(
@@ -374,6 +398,15 @@ pub(super) async fn fetch_dropbox_sample_async(
             write_cache_blocking(&cache, &entry, bytes).await?
         }
     };
+    let metadata_cache = cache.clone();
+    let metadata_path = entry.path_lower.clone();
+    let metadata_revision = entry.rev.clone();
+    let cached_metadata = tokio::task::spawn_blocking(move || {
+        metadata_cache.derived_metadata(&metadata_path, metadata_revision.as_deref())
+    })
+    .await
+    .map_err(|error| format!("Derived Metadata lookup task failed: {error}"))?
+    .map_err(|error| format!("Derived Metadata lookup failed: {error}"))?;
     let decode_path = local.clone();
     let decoded = tokio::task::spawn_blocking(move || {
         vibez_audio_io::file_io::decode_audio_file(&decode_path).map_err(|e| e.to_string())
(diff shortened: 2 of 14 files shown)


alexanderwanyoike:

The core idea is right — source tempo must not be a function of Project tempo, and caching one fit per source is the correct shape — but findings 1 and 2 are user-visible regressions in the exact behaviour the PR sets out to fix, and 3 turns a common loop shape into a silent RAW fallback.

## Bugs

1. **fix** — `crates/vibez-ui/src/state/browser_state.rs:488` — A dragged entry is imported at the *selected* entry's tempo. `audition_import_input()` prefills `source_bpm` from `waveform_loop_fit` guarded only by `waveform_source == selected_source`, and `resolve_for_audio` (`state/mod.rs:56`) short-circuits whenever `source_bpm.is_some()`. But `dispatch_drop_for_target` (`app/media_import.rs:39`) calls `audition_import_input()` for whatever source is being dropped, and drag-start never selects (`views_browser.rs:452` binds `on_press` to `BeginPendingBrowserDrag`; only `on_release` selects, and `begin_pending_drag` does not touch `selected_source`). Failure: click entry A (90 BPM), then drag entry B (128 BPM) into the arrangement — B is warped as if it were 90 BPM. Confirmed with a probe test against `BrowserState`. This is a regression: before the PR `audition_import_input` always returned `None` and the fit was computed from the actual audio. Key the cached fit to the source being imported, the way `drag_preview_beats` (`browser_state.rs:562`) already does.

2. **fix** — `crates/vibez-ui/src/state/mod.rs:60` — The import path passes `detected_bpm: None`, so it resolves differently from Audition. Auditions go through `analyse_loop_tempo` (`warp.rs:160`), which feeds `onset::detect_bpm` in as the reference; `resolve_for_audio` hardcodes `None`, falling back to the 120 BPM reference. Measured on a 2-bar 170 BPM loop with no BPM in the filename: `bars: 2, source_bpm: 170` with detection, `bars: 1, source_bpm: 85` without — the file imports at half tempo and double length while the Browser BPM column and the audition both say 170. Note the fix isn't just "pass the detection": `prepare_browser_sample_import` runs on the update thread, so analysis has to be plumbed through the decode task (`decode_and_stage_local_async` currently doesn't analyse at all).

3. **fix** — `crates/vibez-ui/src/warp.rs:151` — The 5% filename-hint veto rejects ordinary loops outright: `if name_hint.is_some_and(|hint| (fit.source_bpm / hint).ln().abs() > 1.05_f64.ln()) { return None }` discards the whole fit rather than falling back to audio evidence. Measured: a 4-bar 128 BPM loop named `Drums_128_Loop.wav` carrying a one-beat tail yields `None`, so WARP silently degrades to RAW — and this file's own doc comment (`warp.rs:170`) says "Real drum loops almost always carry a trailing pickup", making the tail case the common one. The veto also discards an agreeing detection (`fit_loop_tempo(2-bar 124 loop, "Bass_Loop_100_Fm.wav", Some(124.0))` returns `None`). Widen the tolerance, and demote a disagreeing hint to "ignore the hint, use detection" instead of "reject everything".

## Design / DRY

4. **fix** — `crates/vibez-ui/src/app/update_media.rs:487` — `AnalysedBrowserAudio` is destructured immediately so its halves can be passed separately: `install_audition(gen, source, Arc::clone(&audio), loop_fit)` and `install_waveform(source, audio, loop_fit)` take positional args where the audio and its fit must always correspond. Passing `AnalysedBrowserAudio` through makes the pairing unforgeable and removes the destructure-and-resplit at both call sites; the remote path (`update_remote.rs:490`) can construct one too.

5. **fix** — `crates/vibez-ui/src/warp.rs:57` — `LoopGridFit::target_frames` has no production caller: `git grep` finds it only inside `mod tests` (`warp.rs:328`, `:345-346`). It replaced a struct field that was likewise unread outside `warp.rs`, so the PR converted dead data into a dead method plus its own validity guards. Either wire it into the warp render path or drop it and assert the bar count directly.

6. **fix** — `crates/vibez-ui/src/warp.rs:160` — `detect_bpm` runs unconditionally on every preview, including when its result is discarded: `fit_loop_tempo` uses `detected_bpm` only when `name_hint` is `None`, yet `analyse_loop_tempo` always computes it first. `onset::detect_bpm` allocates four full-length `Vec<f32>` plus prefix arrays over the entire file, so clicking a 5-minute stem in the Browser burns hundreds of MB of transient allocation for a value that is either unused (filename has a BPM) or unusable (no bar candidate lands in 60–200 for anything over ~64 s). It's inside `spawn_blocking` so it won't stall the UI, but gating it behind the cheap candidate/hint check is a few lines.

## Nits

7. **note** — `crates/vibez-core/src/onset.rs:170` vs `crates/vibez-ui/src/warp.rs:72` — the detector's octave fold is `[60, 180]` while the fit's plausible band is `[60, 200]`. A genuine 190 BPM loop has its detection folded to 95, which then steers `fit_loop_tempo` toward the half-time candidate; the detection actively argues for the wrong answer rather than abstaining.

8. **note** — `crates/vibez-ui/src/warp.rs:100` — the `bpm` token's preceding neighbour wins over its following one even when explicitly marked: `Loop_90_bpm_128.wav` returns 90. Ambiguous naming, low stakes, but the preference is arbitrary and undocumented.

alexanderwanyoike:
Addressed every review finding in 6fa66e4.

1. Import treatment is now keyed to the actual source being dropped, so a dragged entry cannot inherit the selected entry BPM.
2. Browser decode tasks carry one AnalysedBrowserAudio value through audition and import. Onset analysis stays off the UI thread, and a fast loop imports with the same source tempo used for audition.
3. Filename hints allow normal loop tails. A disagreeing hint is demoted in favour of detected audio evidence instead of forcing RAW.
4. Audio and its loop fit remain paired through BrowserState APIs and remote materialization.
5. The production-unused LoopGridFit target_frames method is removed.
6. Explicit filename evidence and impossible grid candidates skip onset detection.
7. The onset fold range now matches the 60 to 200 BPM fit range, with a 190 BPM regression test.
8. A BPM value after an explicit bpm token wins over an unrelated preceding number.

Local verification passes for the full workspace tests, both strict Clippy jobs, formatting and diff checks. The fresh three-platform CI run is in progress.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge after addressing new finding 1

All three original bugs are genuinely fixed and were re-verified with probe tests against the new code; findings 4–8 are resolved too (6 better than asked: detection is short-circuited when the filename decides, with a test that panics if detection runs). Full suites pass (563 vibez-ui tests, onset 20/20 including the new 190 BPM case). One residual bug on the remote path plus cleanups.

1. Drag-import used the *selected* entry's tempo — **resolved** (`audition_import_input()` became `import_input_for_source(&source)`, `browser_state.rs:487`, guarded on `waveform_source == source`; all three call sites pass the source being imported; probe: dragging B while A is selected now resolves B's own 128 BPM).
2. Import path skipped onset detection — **resolved** (`resolve_for_analysis` consumes the real `analysed.loop_fit`; every local/staged/hydrate/dropbox entry point feeds a real analysis; probe: 2-bar 170 BPM loop imports as 170, not 85). Residual on the remote path: new finding 1.
3. 5% filename-hint veto rejected ordinary loops — **resolved** (`warp.rs:130`: hint is filtered against the candidate set instead of vetoing the result, tolerance widened to 1.10; probes: the pickup-tail loop and the agreeing-detection case both fit now).
4. Split audio/fit params — **resolved** (`install_waveform`/`install_audition` take `AnalysedBrowserAudio`; the pair can't be mismatched in transit).
5. Dead `LoopGridFit::target_frames` — **resolved** (deleted).
6. Unconditional `detect_bpm` — **resolved** (short-circuits on filename hint and on no-plausible-candidate, covering the 5-minute-stem case).
7. Fold range mismatch — **resolved** (`BPM_FOLD_HI` 180 → 200, matching `PLAUSIBLE_LOOP_BPM_MAX`, with test). Blast radius: new finding 3.
8. `bpm` token order — **resolved** (following token wins, documented, tested).

## New findings

### Bugs

N1. **fix** — `crates/vibez-ui/src/app/async_helpers.rs:401` — The remote *import* runs a fresh `detect_bpm` (`analyse_browser_audio_async`) while the remote *audition* builds its fit from cached `materialized.metadata.bpm` (`update_remote.rs:491`), so the two can disagree for the same file. Failure: audition a remote ~190 BPM loop whose cached metadata predates this PR's fold change (`bpm` reads 95), then import it — audition resolves against 95, import against a fresh 190, different bar counts for the same file. `DropboxCache::derived_metadata` (`vibez-dropbox/src/cache.rs:299`) already exposes the cached value and `cache` is in scope at line 361 — reading it instead of recomputing closes the divergence and saves a redundant whole-file onset pass per remote import.

### Design / DRY

N2. **fix** — `crates/vibez-ui/src/warp.rs:167` — `analyse_loop_tempo_with` re-derives the hint-agreement test that `fit_loop_tempo` performs internally at `:130` (two copies of `bpm_hint_from_name` + the `FILENAME_TEMPO_TOLERANCE.ln()` comparison, same file). They agree today only because the selection rule picks the candidate nearest the hint; the predicates are different statements ("any candidate agrees" vs "the chosen fit agrees"), so a future selection change desyncs the short-circuit silently. Have `fit_loop_tempo` report whether the hint decided.

N3. **note** (PR body) — `crates/vibez-core/src/onset.rs:171` — `BPM_FOLD_HI` 180 → 200 is a vibez-core change with three consumers outside Browser warp, one of which *persists* the value (`DerivedMetadata.bpm`). The change is narrow and correct, but an existing project with a ~190 BPM clip will auto-warp differently after this merge, and already-cached remote entries keep the old halved value while new ones get the new value. Deserves a line in the PR body.

N4. **fix** — `crates/vibez-ui/src/app/media_import.rs:74` and `:105` — the decode-then-analyse async block is byte-identical in both staged arms, with the hydrate arm and `decode_analyse_and_stage_local_async` as two more variations. One `decode_and_analyse_async(path, name)` helper covers both staged arms outright.

### Nits

N5. **note** — `crates/vibez-ui/src/warp.rs:126` — `fit_loop_tempo` now heap-allocates a `Vec` for five fixed candidates to iterate twice; the previous version was allocation-free via `LOOP_GRID_BARS.map(...)`. Runs per browser click; an array avoids it.

N6. **fix** (one comment) — `crates/vibez-ui/src/warp.rs:73` vs `crates/vibez-core/src/onset.rs:171` — `PLAUSIBLE_LOOP_BPM_MAX` and `BPM_FOLD_HI` must agree by design but live in different crates with nothing recording the invariant; a cross-referencing comment stops the next tuning pass from reintroducing prior finding 7.

alexanderwanyoike:
Addressed every re-review finding in 5c5a7e6.

- Remote imports now reuse revision-matched cached tempo evidence, exactly like audition. A regression test reproduces the old 95 BPM audition versus 190 BPM import disagreement.
- Filename-hint authority is returned by the fitter instead of re-derived with a second predicate.
- Duplicate staged decode-and-analyse blocks now share one helper.
- The five fixed grid candidates use an array with no heap allocation.
- Cross-comments record the detector and fitter maximum-BPM invariant.
- The PR body now documents the 180 to 200 BPM fold change and its cache compatibility effect.

Full workspace tests, strict Clippy, formatting and diff checks pass.

alexanderwanyoike:
Final verification is green on all eight Linux, macOS and Windows CI jobs.

alexanderwanyoike:

All six round-2 findings are resolved in 5c5a7e6; no new findings. CI green (Linux/macOS/Windows).

- N1 (remote import re-detected while audition used cached metadata) — **resolved**: `fetch_dropbox_sample_async` now reads `DropboxCache::derived_metadata` for the entry's revision and feeds `metadata.bpm` into the same `fit_loop_tempo` construction the audition path uses (`update_remote.rs:492`), skipping the redundant onset pass. Verified both paths now build the fit from identical evidence, including the metadata-present-but-`bpm: None` case. The new regression test pins the exact round-2 failure scenario (cached 95 wins over a fresh 190 detection).
- N2 (duplicated hint-agreement predicate) — **resolved**: `fit_loop_tempo_decision` reports `filename_hint_decided` and `analyse_loop_tempo_with` consumes it; the predicate now has one home, so a future selection-rule change cannot desync the short-circuit.
- N3 (fold-change blast radius in PR body) — **resolved**: the body documents the 180→200 move, that stale cached remote analysis keeps its half-time value until rematerialised, and that audition/import intentionally share the revision-matched cached value.
- N4 (byte-identical staged decode-analyse arms) — **resolved**: both arms call the new `decode_and_analyse_async` helper.
- N5 (heap-allocated candidate Vec) — **resolved**: back to allocation-free `LOOP_GRID_BARS.map` with `Option` slots and `flatten()`.
- N6 (unrecorded cross-crate invariant) — **resolved**: cross-referencing comments on both `BPM_FOLD_HI` and `PLAUSIBLE_LOOP_BPM_MAX`.

PR #189, opened 2026-08-23
feat: add Audio Clip Inspector V1

alexanderwanyoike:
## What changed

- turns the existing Clip detail area into one compact Audio Clip Inspector for Arrange and Section construction
- adds nondestructive Gain, Source BPM, Warp, Source Start/End, Loop Start/End, and Transpose controls
- applies Clip Gain before the Track channel strip in live playback, resident Sections, Bounce, and export
- renders Transpose and combined Warp/Transpose through Signalsmith away from the realtime thread
- persists the new Clip properties in Project Format V1 with neutral defaults for older projects
- keeps edits location-safe across async work, one Undo step per committed field, and preserves properties through move, split, duplicate, paste, capture, quantize, replay, and tempo-follow paths

## Verification

- cargo check --workspace --all-targets
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --workspace --all-targets
- focused regression coverage for Inspector validation, resident engine commands, Section projection, live/offline gain, persistence, Undo snapshots, Warp round trips, and duration-preserving pitch shift

## Scope

Fades, reverse, editable transient/Warp markers, and slicing remain separate roadmap releases. Native producer QA is the final open acceptance item.


alexanderwanyoike:

The architecture is right: value objects in vibez-core, lock-free engine commands consistent with the drop invariant, gain threaded through live playback, Sections, offline render, capture, and replay with tests, and a load path that covers warped+transposed, transpose-only, and neutral clips. But the inspector editing flow has five real bugs (two of them async races), and the PR pushes two files over the hard 1,000-line limit.

## Bugs

1. **fix** — `crates/vibez-ui/src/domains/arrangement/media_ops.rs:281` — Loop Start can never be committed on a clip whose `loop_end` is still 0. The `LoopStart` arm validates against the *current* `clip.loop_end` (`new_start >= new_end` rejects everything when `loop_end == 0`), and import/bounce creation paths leave `loop_end: 0`. The user gets "Loop bounds must stay ordered inside 0.000 to N s" for every value until they discover that toggling LOOP ON first seeds the bounds. Symmetrically, Loop End is rejected on any clip with `source_offset > 0` and `loop_start == 0` by the `new_start < clip.source_offset` term. Neither Loop field has a test (`domains/arrangement/tests.rs` has zero `LoopStart`/`LoopEnd` references). Validate the pair against what the pair *will be*, seeding the untouched half sensibly, and add tests.

2. **fix** — `crates/vibez-ui/src/widgets/effect_knob.rs:494` — scroll-wheeling the Transpose knob fires one full offline time-stretch and one whole-project undo snapshot per notch. The wheel arm calls `immediate_value_message`, which routes through `SetAudioClipInspectorValue` → `commit_audio_clip_inspector_field` → a `ClipTransposeRenderRequest`; it carries no `UndoGestureId` (unlike the drag arm at :465) and `SetAudioClipInspectorValue` is not in the deferred-edit list at `app/update.rs:91`. Scrolling 0 → +12 st spawns roughly a dozen concurrent whole-clip phase-vocoder renders and leaves a dozen undo steps. The drag path (preview + render-on-release) is the model; wheel should debounce/commit-on-settle or at least share one undo gesture.

3. **fix** — `crates/vibez-ui/src/app/media_import.rs:564` with the guard at `media_ops.rs:123` — a Transpose committed while a warp render is in flight is silently dropped, permanently. The warp dispatch snapshots `transpose_semitones` at request time; if the user commits Transpose +5 during the render, the warp lands first (sets `clip.warped = true`, audio rendered at the old transpose), then the transpose result fails the `clip.warped != success.expected_warped` guard and is discarded. End state: inspector reads +5 st, audio is unpitched, and nothing ever re-renders. The stale-result guard is correct to discard the audio, but a discarded-because-stale result needs to requeue a render from current state instead of returning silently.

4. **fix** — `crates/vibez-ui/src/domains/arrangement/media_ops.rs:685-701` — clear-warp's geometry snapshot silently overwrites edits committed while its render is in flight. `apply_clear_clip_warp` captures source/loop geometry into the render request; on completion `apply_clip_transpose_success` writes it back guarded only by `transpose` and `warped` equality, neither of which changes when the user resizes the clip, commits a Source/Loop bound, or quantizes during the window. Repro: clear Warp on a long clip, drag its right edge shorter while the render runs; the clip snaps back to its old length when the render lands.

5. **fix** — `crates/vibez-ui/src/state/mod.rs:375` (read at `views_detail.rs:1048`) — uncommitted inspector text never expires. Entries in `audio_clip_inspector_edits` are removed only on successful commit (`media_ops.rs:168`) and BPM detection (:806); nothing clears them on blur, Esc, selection change, clip deletion, undo, or project load. Type into SOURCE END, click away without Enter, resize the clip elsewhere, reselect: the field shows the stale text and a later Enter applies a value computed against pre-resize state. Entries for deleted clips accumulate for the session.

## Design / DRY

6. **fix** — `crates/vibez-ui/src/app/views_detail.rs:1019` — the warp readout no longer tells the truth about the rendered tempo. `warp_target` shows `format!("TO {:.0} BPM", self.state.transport.bpm)` (the project tempo); `clip.warped_to_bpm` has zero references left in the file, and the PR deletes the stale-warp badge and the "OUT OF TEMPO" warning whose removed code carried an explicit comment defending it. A clip warped to 120 in a project now at 128 displays as aligned. Read `clip.warped_to_bpm` for the label and keep (or consciously re-decide) the mismatch warning.

7. **fix** — `crates/vibez-ui/src/domains/arrangement/mod.rs:1000-1006` — the knob commit path round-trips f32 through a string and pushes a type error to runtime. `SetAudioClipInspectorValue` formats the value to text (duplicating the knob's own formatting at `effect_knob.rs:203`), inserts it into the edit map, and immediately re-parses it; the `_ =>` arm answers a constructible `SourceStart` knob message with the status "This Audio Clip control is not rotary". A rotary-only field enum for `KnobTarget::AudioClip` removes both the round-trip and the dead branch.

8. **fix** — file-size rule: this PR pushes `views_detail.rs` from 890 to 1,261 lines and `domains/arrangement/media_ops.rs` from 967 to 1,228, both crossing the repo's hard 1,000-line limit; `update.rs` (1,042 → 1,122) and `message.rs` (1,011 → 1,037) were already over on dev and grew further. The inspector view and `commit_audio_clip_inspector_field` are each cohesive enough to lift into their own modules; do that split on this PR for the two files it pushed over.

## Nits

9. **fix** — `media_ops.rs:168` — the edit-map entry is removed *before* parsing, so rejected input is discarded rather than left for correction: type `abc` into Gain and the field snaps back to the committed value alongside the error message.

10. **fix** — `media_ops.rs:305` — Transpose parses as `i8`, so `0.5` or `-1.5` yields a hard range error (the message prints the ±48 bounds), while the knob writing the same field always rounds. Parse as f64 and round to match the knob's contract.

11. **note** — `crates/vibez-dsp/src/time_stretch.rs:131` — when Signalsmith's `exact` declines, the `stretch_to` fallback resamples without applying the semitone shift; for a pure transpose render (target == source frames) that fallback is an identity, so a committed Transpose silently becomes a no-op with no surfaced warning. Rare path; worth at least a warning in the result.

## Clean

Verified clean: `ClipGainDb`/`ClipTranspose` clamp on deserialize with legacy-project defaults; engine drain handlers are allocation-free and follow the existing Arc-drop invariant; gain parity is tested in live playback, Section projection, and offline master render; `TimelineLocation` is threaded through quantize, BPM detect, warp, and transpose with the arrange/selected-section/non-selected-section cases handled separately; clip properties survive move, split, duplicate, paste, capture, quantize, and replay; no panic paths found in the new code.

alexanderwanyoike:
Addressed the full review in `5839fc3`.

- Loop fields now seed the untouched bound from the current source range.
- Transpose wheel input is revision-debounced, so one settled gesture produces one render and one undo step.
- Transpose and clear-Warp renders now carry audio and geometry expectations. Stale results are rejected and requeued from current clip state.
- Inspector drafts survive validation errors, then clear on real blur, Escape, selection changes, clip removal, geometry replacement, undo, and project load.
- Audio clip knobs use a rotary-only field type and commit typed values without formatting and reparsing.
- Fractional Transpose text rounds to the same semitone contract as the knob.
- The Warp readout shows `warped_to_bpm` and restores the `OUT OF TEMPO` warning.
- Signalsmith transpose fallback now reaches the UI as a warning instead of silently reporting success.
- Split the Audio Clip view and inspector editing lifecycle into dedicated modules. `views_detail.rs` is 710 lines and `media_ops.rs` is 973 lines.

Regression coverage includes both async races, uninitialised loop bounds, rejected draft retention, fractional Transpose, wheel debounce revisions, and outside-click blur.

Verified with the full workspace test suite, 587 UI tests, strict workspace Clippy with warnings denied, formatting, and `git diff --check`.

alexanderwanyoike:

All eleven findings are resolved in 5839fc3, several more thoroughly than asked. CI is green on all platforms. No new bugs found in the fix commit.

1. Loop Start uncommittable on fresh clips — **resolved**: the Loop arms now seed the untouched half of the pair (`audio_clip_inspector.rs`, LoopStart falls back to `source_end`, LoopEnd to `source_offset`) so a first-time commit works without toggling LOOP ON first.
2. Wheel transpose fired a render + undo snapshot per notch — **resolved**: wheel routes through `PreviewAudioClipRotaryValue` (draft + monotonic per-clip revision), a 180 ms settle task commits only if the draft and revision are still current, and drag previews via `InputChanged` with one Submit on release. Gain stays immediate, which is real-time safe and matches every other knob target's wheel convention.
3. Transpose dropped during an in-flight warp render — **resolved**: `ClipTransposeSuccess` now carries `expected_audio` (Arc identity) and stale results *requeue* a fresh render from current state with a status message instead of returning silently.
4. Clear-warp geometry snapshot overwrote concurrent edits — **resolved**: the request carries `expected_geometry`; a resize during the render makes the result stale and requeues rather than snapping the clip back.
5. Uncommitted inspector text never expired — **resolved**: selection change discards all drafts, an orphan sweep after every arrangement message drops entries for deleted clips, geometry-replacing renders discard per-clip, and `DiscardAudioClipInspectorEdit` exists for explicit cancel.
6. Warp readout showed the project tempo — **resolved**: `views_audio_clip.rs:469` reads `clip.warped_to_bpm` and the OUT OF TEMPO warning is restored.
7. Knob f32-through-string round-trip + "not rotary" runtime arm — **resolved**: typed `AudioClipRotaryField` (Gain | Transpose) makes a knob emitting a bounds edit unrepresentable; the dead arm is gone.
8. File-size rule — **resolved**: `views_detail.rs` 1,261 → 710 + `views_audio_clip.rs` 601; `media_ops.rs` 1,228 → 973 + `audio_clip_inspector.rs` 416. All new-code files are under 1,000. (`update.rs`/`message.rs` remain slightly over, as they were on dev.)
9. Rejected input discarded — **resolved**: the draft is read with `get`, removed only after a successful parse and apply.
10. Transpose `i8` parse — **resolved**: parses f64, rounds, range-checks ±48.
11. Signalsmith fallback silently dropped the shift — **resolved**: `pitch_preserving_stretch_transposed_checked` reports the fallback and the result surfaces "Transpose was not applied" in the status.

## Notes

N1. **note** (cross-PR, on rebase) — the inspector's Loop commit still validates bounds inline. #191 (via #190) introduces `UiClip::set_loop_region` as the canonical validator with the same range; when this PR rebases over the loop-marker stack, route the Loop arm through it so the predicate keeps one home.

N2. **note** — `update.rs` `CommitAudioClipTransposeAfterDelay` drops the commit when the active timeline location changed during the 180 ms window, leaving the uncommitted draft in the edits map until the next selection change clears it. Narrow and self-healing; fine to leave.

PR #190, opened 2026-08-23
fix: make MIDI clip loops follow explicit clip bounds

alexanderwanyoike:
## What changed

- activating Loop now starts from the complete MIDI Clip instead of reusing stale bounds or guessing from note content
- extending a non-looping MIDI Clip no longer silently enables a shorter inferred loop
- adds the two-bar stale-bound regression that reproduced authored notes overlapping a ghost repeat

## Why

A previously resized Clip could retain a valid-looking one-bar loop region. Enabling Loop on the two-bar Clip reused that stale region, so the second bar remained authored while the first bar repeated over it. Loop boundaries are now explicit producer intent.

## Verification

- `cargo test --workspace --all-targets`
- `cargo clippy -p vibez-ui --all-targets --all-features -- -D warnings`


alexanderwanyoike:
# mini-code-review — verdict: merge after fixing finding 1, and only together with #191

The core semantic change is right and the three regression tests genuinely pin the bug (each fails on dev; suite passes on the branch, 572 tests). One real bug in the multi-select path, and two consistency decisions that belong to this stack.

## Bugs

1. **fix** — `crates/vibez-ui/src/domains/arrangement/ops.rs:393` — the multi-clip toggle rewrites the loop region of clips that were already looping. `enable` is computed once for the whole selection as "any selected clip is not looping" (ops.rs:350), then the new unconditional `if enable { start = 0.0; end = duration }` runs for every note clip, including one whose `loop_enabled` was already true and which this gesture does not actually change. Select a MIDI clip carrying a hand-set 1-bar loop plus a second non-looping clip, press L: the first clip's region silently jumps to the full clip length. Latent today, guaranteed once #191 ships marker dragging. The audio arm directly above is immune (its reset stays guarded by `loop_end <= loop_start`). Gate on the transition: `if enable && !was_enabled`.

## Design / DRY

2. **note** — merged alone, this PR is a user-visible regression: `SetNoteClipLoopRegion` has no producer anywhere in the view layer on this branch, so every MIDI loop region is exactly `0..duration_beats`, which the engine renders identically to looping off (each note once, no boundary flush). Extend-then-loop, which worked on dev via content inference, now yields silence in the added bars with no in-app way to shorten the region. Intended, since #191 supplies the markers, but it means this PR must land with or immediately before #191, never alone. Relatedly, the unclamped write in the `SetNoteClipLoopRegion` handler (stores bounds verbatim, so an inverted region would silently disable engine looping) is fixed by #191's validation — do not double-fix it here.

3. **fix (in this stack)** — `crates/vibez-ui/src/domains/arrangement/media_ops.rs:301` — audio and MIDI now disagree about extending a clip past its content: the audio resize path still auto-enables the loop when dragged past its source length, which is exactly the inference this PR removes for MIDI. Same drag on the same timeline, opposite outcomes: the audio clip repeats, the MIDI clip goes silent. The audio toggle also preserves a stale region on re-enable (it re-seeds only when `loop_end == 0`), the exact stale-bounds behaviour this PR eliminates for MIDI. If "loop bounds are explicit producer intent" is the rule, the audio paths are now the ones out of step; align them somewhere in this stack.

4. **fix** — "reset the loop to the whole clip" is written out four times (ops.rs:395, piano_roll.rs:219, media_ops.rs:204, media_ops.rs:945) and the shrink-clamp twice (ops.rs:486, piano_roll.rs:681). The concrete cost is finding 1: the missing transition guard exists in exactly one of the four copies. A shared `UiNoteClip::enable_loop_over_clip()` / `clamp_loop_to_duration()` makes it impossible to get wrong in only one place.

5. **note** (decision, not a fix) — toggling Loop off and back on now discards a deliberately dragged region, since the code cannot distinguish "stale after a resize" from "what the user drew". It is undoable and marks the project dirty, but it is the behaviour #191 users will hit most often; worth a conscious product decision recorded somewhere.

## Nits

6. **note** — `widgets/piano_roll/draw.rs:254` — with the region always equal to the clip on enable, the loop shading tints the whole roll and the boundary lines sit at beat 0 and the right edge; the shading marks nothing until the region is shortened.

7. **note** — `domains/piano_roll.rs:1057` — `toggle_loop_defaults_region_to_the_complete_clip` is now fully subsumed by the new `enabling_loop_replaces_stale_bounds_with_the_current_clip_length`.

alexanderwanyoike:
Review addressed in 636c765. Mixed selections now preserve regions on clips that were already looping, audio and MIDI both use explicit loop activation instead of resize inference, and the duplicated reset/clamp rules live on the clip types. Added regressions for mixed selections and audio resize. Full workspace tests, format, and strict Clippy pass; all 8 GitHub checks are green.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge as-is, still only together with #191

All findings resolved in 636c765; CI green.

1. Multi-select toggle rewrote already-looping clips' regions — **resolved**: both arms now gate the reset on the transition (`enable && !was_enabled`), with new regression tests.
2. Standalone-merge regression — unchanged by design; the stack ordering note stands: land with or immediately before #191.
3. Audio/MIDI semantic asymmetry — **resolved**: the audio resize auto-enable inference is removed (resize now clamps an enabled loop via `clamp_loop_to_clip`), and the audio toggle resets to the full clip on enable exactly like MIDI. The two clip types now share one product rule: loop bounds are explicit producer intent.
4. Four copies of the reset + two clamps — **resolved**: shared `reset_loop_region_to_clip` / `enable_loop_over_clip` / `clamp_loop_to_duration` / `clamp_loop_to_clip` on `UiClip` and `UiNoteClip`; all call sites converted.
5. Toggle-discards-authored-region decision — settled implicitly by 3: reset-on-enable is now the consistent rule everywhere. Worth one line in the notes repo if it ever surprises a dogfood session.
6/7. Nits — the subsumed test was deleted; the loop-shading cosmetic note stands as informational.

PR #191, opened 2026-08-23
feat: edit Clip loops with draggable markers

alexanderwanyoike:
## What changed

- draws Ableton-style loop braces with directional start and end handles above the MIDI piano roll and Audio waveform
- makes both handles directly draggable with larger invisible hit targets and horizontal resize cursors
- snaps MIDI bounds to the active grid while Audio bounds edit source frames
- groups every marker drag into one Undo step
- prevents crossed, inverted, out-of-Clip MIDI bounds and out-of-source Audio bounds
- shares the marker geometry and visual language across both editors

## Native QA

Verified on an isolated native Vibez display with a disposable project:

- MIDI end marker extended an 8-beat loop to 12 beats and the repeat boundary moved correctly
- Audio end marker shortened the source loop and the waveform repetitions updated immediately
- both edits dirtied and autosaved only the disposable project

## Verification

- `cargo test --workspace --all-targets`

## Stack

This PR targets #190 because the marker interaction relies on the corrected MIDI loop activation semantics.


alexanderwanyoike:
# mini-code-review — verdict: merge after fixing findings 1-3

The interaction skeleton is solid: one undo gesture per drag (verified through `UndoHistory::push_edit` coalescing), consistent local-space coordinates with no scroll/zoom errors, no click stealing from note editing or the ruler, and no new per-frame allocation beyond the surrounding code's idiom. The bugs are all in the audio-side drag math and the no-op handling. Suite passes on the branch (578 tests).

## Bugs

1. **fix** — `crates/vibez-ui/src/widgets/audio_clip_detail.rs:329` and `:338` — an audio loop collapses to a single sample when a handle overshoots, and the end handle then becomes ungrabbable. The audio drag uses a one-frame minimum (`fixed_end.saturating_sub(1)` / `fixed_start.saturating_add(1)`) instead of a musical minimum. Drag the start brace past the end brace and you land on `loop_start == loop_end - 1`: the engine loops one frame (an audible buzz) and both braces render on the same pixel, where the hit-test tie-break (`start_distance <= end_distance`) always returns Start, so only Undo or dragging start back recovers (`ToggleClipLoop` re-seeds only when `loop_end == 0`). The MIDI path gets this right with a one-grid-cell `min_length` (`piano_roll/mod.rs:650`); use the same idea in frames.

2. **fix** — `audio_clip_detail.rs:55`/`:63` — a loop region extending past the visible clip window is silently truncated by the first pixel of a drag. `source_to_x` and `x_to_source` both map through `[source_offset, source_offset + duration_samples]`, so a `loop_end` beyond that window draws pinned at the right edge and the first `CursorMoved` maps that edge back to the window end. Reachable: `op_resize_audio_clip` (`media_ops.rs:298`) shrinks `clip.duration` without clamping loop bounds, so shrinking a clip in the arrangement then nudging the end handle jumps the loop from, say, 200k frames to 40k with no visual cue. The same mapping also makes the source tail between the window end and `num_frames` unreachable by drag even though the domain validator accepts it.

3. **fix** — `app/update.rs:98` with both widgets — every marker drag dirties the project and pushes an undo snapshot even when nothing changed, and streams identical engine commands per mouse-move. Neither widget compares the computed bounds against the current pair before emitting, and `should_mark_dirty`/`push_undo_snapshot` run before validation. Jiggle a brace two pixels inside one 16th cell: values are identical, but the project is dirty and an undo entry undoes nothing. Early-return in the widget when the new pair equals the current pair; that also stops the ~60/s stream of identical `SetNoteClipLoop`/`SetClipLoop` commands into the engine ring buffer.

## Design / DRY

4. **fix** — `widgets/clip_loop_markers.rs` — the "shared" module shares only the two cheapest pieces (hit test + brace drawing, 91 lines) while each call site re-declares the drag state (`fixed_end`/`fixed_start` + `UndoGestureId`), re-implements the clamp shape, and re-wires press/move/release. The clamp is exactly where bug 1 diverged between the two editors. A shared `LoopDrag` enum with a `resolve(pointer, min_length, max) -> (start, end)` covers both, leaving each call site only unit conversion and message construction.

5. **note** (decision) — `audio_clip_detail.rs:63` — audio loop bounds ignore the snap grid entirely while MIDI snaps; the PR body declares this ("Audio bounds edit source frames"), but on an 800px canvas over a 4-bar clip every hand-set loop lands tens of milliseconds off the bar and drifts audibly on each repeat. The detail panel already has a quantize row and the app a global grid, so snapping is available if the product answer is "musical by default, free with a modifier".

6. **fix** — `audio_clip_detail.rs:130-163` vs `:239` — the canvas now marks the same loop twice: the pre-existing full-height start line and repeat-boundary lines, plus the new accent brace at the same x positions. Two visual languages for one value; drop the old start line or unify the colour treatment.

7. **fix** — `clip_loop_markers.rs:6` vs `piano_roll/mod.rs:180` — the shared module's `MARKER_RAIL_HEIGHT` (18.0) is used by the audio widget while the piano roll passes `RULER_HEIGHT` (20.0), so the grab strip differs between two editors meant to feel identical.

8. **note** (cross-PR, coordinate with #189) — `domains/arrangement/mod.rs:403` — this PR's validator accepts `loop_start >= 0 && loop_end <= source_frames`, while #189's inspector Loop fields require bounds inside `[source_offset, source_offset + duration]`. When both PRs merge, the same field pair has two different legal ranges depending on which control set it (markers accept a region the inspector then refuses to re-commit). Align on one predicate in one place before both land.

## Nits

9. **fix** — `widgets/piano_roll/draw.rs:561` — the brace paints directly over the ruler's bar numbers (labels at y=3, size 10; brace fills y=1..10). The audio widget lays a 0.92-alpha surface rail first; the piano roll should do the same or move the labels.

10. **fix** — `piano_roll/mod.rs:355` — Draw mode returns the Crosshair cursor for the whole canvas before the marker hit test, but the press handler starts a loop drag regardless of edit mode, so a live handle advertises the note-drawing cursor.

11. **fix** — `audio_clip_detail.rs:340` — the End branch emits `loop_start = fixed_start.min(source_end - 1)` instead of the captured `fixed_start`, so an end drag can silently move the start on a clip whose source got shorter (relinked media). Send `fixed_start`.

12. **note** — file-size rule: nothing newly crosses 1,000 lines, but three already-over files grow: `domains/piano_roll.rs` 1,186 → 1,235, `domains/arrangement/mod.rs` 1,029 → 1,034, and `widgets/piano_roll/mod.rs` 1,026 → 1,216 (+190, the largest growth in the PR). The loop-drag arm in `widgets/piano_roll/mod.rs` is a self-contained slice; splitting it out alongside `draw.rs` would take that file back under the limit.

alexanderwanyoike:
Review addressed in b2d9edc. The two editors now share one LoopDrag resolver; audio uses a musical minimum and project-grid snapping, loop bounds stay inside the visible clip, unchanged drags emit no edit, duplicate marker lines are gone, rail geometry/cursors match, and the piano-roll main module is back under 1,000 lines. Added regressions for overshoot, legacy out-of-window bounds, no-op drags, cursor feedback, and domain no-ops. Full workspace tests (585 UI tests), format, and strict Clippy pass locally.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge as-is once CI has run on this head

All twelve findings are resolved in b2d9edc (with the #190 base merged in). One logistics note: GitHub reports no checks on this branch head because the PR targets `fix/midi-clip-loop-region`; CI will run when it retargets dev after #190 merges — confirm it goes green before merging.

1. One-frame minimum collapsed the audio loop and stranded the end handle — **resolved**: shared `LoopDrag::resolve` enforces a one-grid-cell minimum in frames (0.01 beat unsnapped), with an overshoot regression test.
2. Out-of-window region truncated by the first drag pixel — **resolved**: `#190`'s `clamp_loop_to_clip` stops the state from being created on resize, the drag max is now explicitly `min(window, source remainder)`, and a legacy out-of-range region is deliberately repaired on drag with a test pinning the behaviour.
3. No-op drags dirtied the project and streamed engine commands — **resolved**: equality early-return in both widgets plus `set_loop_region` returning false on identical or invalid bounds (so nothing marks dirty and no command is sent), with tests on both layers.
4. Drag logic duplicated across editors — **resolved**: `LoopDrag` (begin/resolve/undo_gesture) lives in `clip_loop_markers.rs`; each call site keeps only unit conversion and message construction. The piano-roll wiring moved to `widgets/piano_roll/loop_drag.rs`.
5. Audio bounds ignored the snap grid — **resolved beyond the ask**: the audio widget now takes the transport BPM + grid config and snaps `x_to_local_frame` through the same `snap_beat` the MIDI editor uses.
6. Same loop marked twice on the audio canvas — **resolved**: the redundant full-height start line is gone and the first repeat-boundary line moved off the brace-end position.
7. 18 px vs 20 px rails — **resolved**: `MARKER_RAIL_HEIGHT` is 20.0 and both editors pass it.
8. Validator divergence with #189's inspector — **resolved on this side**: `UiClip::set_loop_region` / `UiNoteClip::set_loop_region` are the single validators with the inspector's range (`source_offset ..= min(source_offset + duration, num_frames)`); #189 should route its Loop commit through them on rebase (noted there).
9. Brace over bar numbers — **resolved**: the brace draws before the ruler text so labels stay legible, with a comment stating the ordering constraint.
10. Draw-mode cursor over a live handle — **resolved**: the marker hit-test runs before the Draw-mode crosshair return.
11. End drag rewrote `loop_start` — **resolved**: `resolve` returns the captured `fixed_start` unchanged.
12. File-size growth — **resolved**: `widgets/piano_roll/mod.rs` split to 976 lines via `loop_drag.rs` (77) and `tests.rs` (240). `domains/piano_roll.rs` remains over from dev (1,220), unchanged status.

PR #192, opened 2026-08-24
feat(audio): add bars and beats to the Clip waveform

alexanderwanyoike:
## What changed

- adds a musical ruler to the Audio Clip waveform
- draws bar, beat, and subdivision grid lines from project tempo
- keeps the ruler aligned with the same snap grid used by marker dragging
- reserves the ruler rail above the waveform so audio and MIDI editors use the same visual language

## Stack

This PR is based on `test/pr-189-191`, the temporary integration branch containing reviewed PRs #189, #190, and #191. It should be retargeted to `dev` after that stack lands.

## Verification

- Audio Clip widget tests pass
- strict Clippy passes
- formatting passes

PR #193, opened 2026-08-24
feat(clips): add independent Start markers

alexanderwanyoike:
## What changed

- adds an independent Start marker to audio and MIDI clips
- keeps the default at Start = Loop Start while allowing intros and pickups before the repeating region
- gives Start its own labelled handle row so it remains draggable when it overlaps Loop Start
- threads Start through playback, offline render, project persistence, warp, copy, split, capture and section playback
- keeps Source In and Source Out separate in the audio inspector
- makes older projects inherit Start from the existing source offset or loop start

## Playback contract

Playback begins at Start. On the first pass it continues to Loop End, then subsequent passes return to Loop Start. A non-looping audio clip stops reading at Source Out.

## Verification

- `cargo test --workspace`
- `git diff --check`

## Stack

This PR is stacked on #192 so the combined build contains the new audio waveform ruler and the three clip markers.


alexanderwanyoike:
# mini-code-review — verdict: merge after fixing findings 1 and 4; schedule finding 2 as an immediate follow-up card before transient/warp markers

The behaviour is correct: I traced the play-position to source-position mapping in both the audio path (`playback_source.rs`) and the MIDI groove path (`groove.rs`) and they implement the same contract with equivalent modulo formulas (including the pickup case, the swing effective-position, and the boundary flush), the migration defaults match where playback previously began for each clip type, split/capture reset the marker sensibly, and the marker widgets separate the Start row from the loop-handle row so overlap stays draggable. Tests cover the contract at the groove, mixer, engine, domain, and widget layers. The concern this PR raises is structural, and it is real; it is named precisely in finding 2.

## Bugs

1. **fix** — `crates/vibez-ui/src/widgets/timeline/clips_draw.rs:445` — the arrangement timeline's ghost-note preview inside loop repeats was deleted rather than remapped for the Start marker. The repeat boundary lines survive (correctly shifted by `start_marker_beats`), but producers lose the visual of *which notes* repeat in the Arrange view, and the PR body does not declare the removal. The helper that can power it correctly landed in this same PR: `UiNoteClip::note_occurrences` (capture.rs already consumes it). Restore the ghost blocks on top of `note_occurrences`, or state the removal as intended in the PR body.

## Design / DRY

2. **fix (as an immediately-following card, before outcomes 8-9)** — the clip playback timeline has no owning abstraction, and this PR makes the cost measurable: one new clip property touched 62 files. The play-time to source-time mapping is now hand-encoded five times across two crates: the audio sample mapper (`playback_source.rs:342`), three sites in the MIDI groove path (`groove.rs`: occurrence mapping, swing effective-position, boundary flush), and the UI's `UiNoteClip::note_occurrences` (`ui_types.rs:309`). All five agree today (verified), but nothing but review discipline keeps them agreeing. On top of that, the marker's legal-window rule (default + clamp) is inlined at roughly eight more sites: `project_io.rs` twice, `project_sections.rs` twice, `render.rs` twice (see finding 3), `media_ops.rs` resize twice, `domains/piano_roll.rs` resize three times, beside the proper `set_start_marker` validators. Recommendation: extract a `ClipTimeline` value object into vibez-core owning `{start, loop_start, loop_end, duration}` with `source_at(play_pos)`, `occurrences_of(source_pos)`, and `clamp_start(...)`, in frame and beat flavours sharing property tests; engine and UI both consume it. This is a mechanical consolidation, not a rewrite, and it is a prerequisite for outcome 9's piecewise time map, which cannot be hand-inlined in five places.

3. **fix** — `crates/vibez-engine/src/render.rs:252` and `:276` — the offline render applies the legacy default (`unwrap_or(source_offset)` / `unwrap_or(loop_start_beats)`) *without* the clamp that `project_io.rs:783` applies on load, so the same field has two migration rules. In-session bounce inputs are already validated so this is latent, but any path that renders a loaded project without passing through the UI clamp (make_demo, future headless export) diverges from live playback. One shared migration helper on `ClipInfo`/`NoteClipInfo` closes it (and folds naturally into finding 2's value object).

4. **fix** — `crates/vibez-ui/src/domains/piano_roll.rs` is at 1,321 lines, over the 1,000 hard limit and grown roughly 100 lines by this PR (it was already flagged on the marker PR). The Start/loop marker message arms are a cohesive slice to lift out, mirroring the `audio_clip_inspector.rs` split from #189.

## Nits

5. **note** — the "minimum gap below Loop End / Source Out" invariant is expressed three different ways: 1 frame in `project_io.rs` and the audio widget, 0.01 beats in the MIDI clamps, one grid cell in the drag paths. Harmless today, but it belongs to finding 2's value object as a single definition.

6. **note** — no CI has run on either stacked branch (`feat/audio-clip-ruler`, `feat/audio-clip-start-marker`); the author's local runs pass, but confirm checks go green when the stack retargets dev before merging.

alexanderwanyoike:
All review findings are now addressed in this branch. The duplicated clip timing geometry has also been followed up immediately in #194, which centralizes the Start and loop rules for audio frames and MIDI beats without adding allocations to the audio path.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge as-is (in stack order; confirm CI when the stack retargets dev)

All four fix findings are resolved in 67e423b + 217731d8, and the architectural finding got a real answer in the stacked #194 (reviewed separately).

1. Ghost-note repeat preview — **resolved**: the timeline clip preview rebuilds note blocks through `note_occurrences`, repeats render at reduced alpha via the new `repeated` flag, and the repeat-region shading is preserved; covered by updated `clips_tests`.
2. Missing playback-timeline abstraction — **addressed by the stacked #194** (`ClipTimeline` value objects in vibez-core); see that review.
3. Two migration rules for the start marker — **resolved**: `ClipInfo::resolved_start_marker` / `NoteClipInfo::resolved_start_marker_beats` are the single defaulting+clamping helpers, with unit tests, consumed by `render.rs`, `project_io.rs`, and `project_sections.rs` (and further unified onto the shared timeline type in #194).
4. `domains/piano_roll.rs` over the 1,000-line limit — **resolved**: split to 797 lines, with the marker arms in `piano_roll/clip_markers.rs` (69) and tests moved to `piano_roll_tests.rs` (478).
5. Epsilon conventions — **resolved beyond the note** in #194: `MIN_START_GAP` is owned by each timeline flavour.
6. CI on the stacked branch — still outstanding by construction; the author's local full-suite runs pass, so this is a merge-time logistics check, not a code concern.

PR #194, opened 2026-08-24
refactor(clips): centralize playback timeline geometry

alexanderwanyoike:
## What changed

- adds shared frame and beat clip timeline value objects in `vibez-core`
- centralizes Start, Loop Start, Loop End, wrapping, occurrence placement and marker clamping
- routes audio playback, MIDI scheduling, waveform previews, capture, split and clipboard mapping through the shared geometry
- preserves real-time safety with allocation-free occurrence iterators that can jump directly to the current audio block
- removes the duplicated engine loop traversal helpers

## Why

This is the direct architectural follow-up from the review of #193. Clip timing rules should live in one domain abstraction rather than being repeated across engine and UI call sites.

## Verification

- `cargo test --workspace` (81 core, 230 engine, 609 UI tests passed)
- `git diff --check`

Stacked on #193.

### DIFF
--- a/crates/vibez-core/src/clip_timeline.rs
+++ b/crates/vibez-core/src/clip_timeline.rs
@@ -1,0 +1,363 @@
+//! Clip-local playback geometry shared by audio, MIDI, engine, and UI.
+
+/// Playback geometry measured in source frames.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FrameClipTimeline {
+    pub start: u64,
+    pub loop_start: u64,
+    pub loop_end: u64,
+    pub duration: u64,
+    pub loop_enabled: bool,
+}
+
+impl FrameClipTimeline {
+    pub const MIN_START_GAP: u64 = 1;
+
+    pub const fn new(
+        start: u64,
+        loop_start: u64,
+        loop_end: u64,
+        duration: u64,
+        loop_enabled: bool,
+    ) -> Self {
+        Self {
+            start,
+            loop_start,
+            loop_end,
+            duration,
+            loop_enabled,
+        }
+    }
+
+    pub fn is_looping(self) -> bool {
+        self.has_loop_region() && self.start < self.loop_end
+    }
+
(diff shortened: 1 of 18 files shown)


alexanderwanyoike:

This is the architectural follow-up the #193 review asked for, and it delivers exactly the right shape. `FrameClipTimeline` / `BeatClipTimeline` in vibez-core now own the play-time to source-time mapping (`source_at`), occurrence placement (`occurrences_of`, `wraps`) as allocation-free iterators with an O(1) `starting_after` jump into the current block, and the start-marker legality rules (`clamp_start`, `accepts_start`, `MIN_START_GAP` per flavour). All five former hand-encodings route through it: the audio sample mapper, the MIDI groove path (whose bespoke `for_each_loop_occurrence`/`for_each_repetition` are deleted outright), the UI's `note_occurrences`, both clip types' validators, and the #193 migration helpers. I verified the translation is semantically faithful at the boundaries that matter: the strict-lower/inclusive-upper visit window (`previous_local < x <= last_local`) survives as `starting_after(...).take_while(...)`, the note-at-loop-end exclusion is enforced inside `occurrences_of`, and the audio modulo matches the old arm including the pickup case. A cross-flavour property test pins frames and beats to the same integer geometry. Real-time safety holds: the hot path constructs a Copy struct and does arithmetic, no allocation.

## Nits

1. **fix (one-liner)** — `crates/vibez-engine/src/playback_source.rs:364` — `clip.timeline()` is constructed inside the per-frame loop; all five fields are loop-invariant, so hoist `let timeline = clip.timeline();` above `for frame in 0..frames`. The optimizer almost certainly does this already, but the codebase's hot-path idiom is to make it visible.
2. **note** — `crates/vibez-engine/src/mixer/groove.rs` — the `end < timeline.loop_end` guard before the note-off occurrence loop restates a condition `occurrences_of` already enforces; harmless, but it re-inlines a piece of the invariant the type now owns.
3. **note** — no CI has run on the stacked branch; confirm checks when the stack retargets dev.

PR #196, opened 2026-08-24
feat: add persisted audio clip fades

alexanderwanyoike:
## Summary

- persist clip-local Fade In and Fade Out lengths with safe defaults for older projects
- apply the same sample-accurate envelope in live Arrange, resident Sections, Bounce, and export
- add precise Inspector controls and preserve valid fades through resize, copy, split, capture, Warp, Transpose, save, and replay
- keep fades anchored to the visible clip edges so loop wraps do not restart them

## Validation

- `cargo test --workspace`


alexanderwanyoike:

Textbook slice. `ClipFades` owns the invariant (fade-in + fade-out never exceed duration, last-edited edge wins, the other shrinks) with the clamping in the constructor; the envelope applies in the one shared render path so live Arrange, resident Sections, Bounce, and export all agree (verified `PreparedPlaybackSource` and `render_offline_inner` both feed the same `EngineClip.fades`); `gain_at` is pure arithmetic and allocation-free, safe for the callback; fades anchor to visible clip edges so loop wraps do not restart them, with a test pinning exactly that; the engine drain defensively re-clamps at every geometry-changing command; and every edit path is covered with the right operator — `for_fragment` for split/copy/capture (edge-ownership tested), `scaled` for warp renders, `clamped_to` for resize and load. Persistence is `skip_serializing_if` with a legacy-default test, and the inspector pair-clamp behaviour is pinned (`0.750 + 0.750` on a 1 s clip resolves to 0.25/0.75). CI is green on all three platforms.

No findings worth raising at this PR's boundary. The one structural pressure this stack creates (`media_ops.rs` approaching the 1,000-line limit) is reported on #198 where it lands at 999.

PR #197, opened 2026-08-24
feat: add timeline fade handles

alexanderwanyoike:
## Summary

- draw Audio Clip fade envelopes and compact handles in Arrange and Section timelines
- keep fade hit targets below the title bar so clip move and trim gestures remain separate
- preview audible fades in real time while dragging
- group every pointer drag into one Undo gesture

## Stack

- depends on #196

## Validation

- `cargo test --workspace`

### DIFF
--- a/crates/vibez-ui/src/app/update.rs
+++ b/crates/vibez-ui/src/app/update.rs
@@ -91,9 +91,7 @@ impl App {
         let owns_project_transaction = self.begin_project_track_deletion_transaction(&message);
         let deferred_arrangement_project_edit = matches!(
             &message,
-            Message::Arrangement(msg)
-                if msg.is_clipboard_project_edit()
-                    || matches!(msg, ArrangementMsg::SubmitAudioClipInspectorField { .. })
+            Message::Arrangement(msg) if msg.defers_project_edit()
         );
         let should_mark_dirty = matches!(
             &message,
@@ -210,9 +208,7 @@ impl App {
                 self.apply_devices_action(action);
             }
             Message::Arrangement(msg) => {
-                let deferred_snapshot = (msg.is_clipboard_project_edit()
-                    || matches!(msg, ArrangementMsg::SubmitAudioClipInspectorField { .. }))
-                .then(|| self.take_snapshot());
+                let deferred_snapshot = msg.defers_project_edit().then(|| self.take_snapshot());
                 let playhead_beats = self.focused_editor_playhead_beats();
                 let samples_per_beat = if self.state.transport.bpm > 0.0 {
                     60.0 * self.state.transport.sample_rate as f64 / self.state.transport.bpm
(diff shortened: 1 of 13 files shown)


alexanderwanyoike:

The interaction design is right: handles only on selected clips, hit band below the title bar so move/trim/seek gestures stay separate, envelope drawn for any faded clip, one `UndoGestureId` per drag with coalescing (tested end to end, including the two-move same-gesture assertion). One recurring bug class and one file-size item.

## Bugs

1. **fix** — `crates/vibez-ui/src/domains/arrangement/mod.rs:517` (`SetAudioClipFade`) — no no-op suppression: every mouse-move during a handle drag rewrites `clip.fades`, sends `SetClipFades` into the engine ring buffer, and marks the project dirty even when the rounded frame value did not change. This is the same class the loop-marker review (#191) fixed with an equality early-return, and the precedent (`set_loop_region` returning `bool`) is sitting in the same file family. Wiggle a handle within one pixel: dirty project, ~60 identical engine commands/s. Suppress before mutating; on the #198 stack the check must run *before* the edge-unlink call so a zero-movement drag also does not dissolve a crossfade link.

## Design / DRY

2. **fix** — `crates/vibez-ui/src/widgets/timeline/clips.rs` — already over the 1,000-line limit on dev (1,012) and this PR grows it by roughly 100 lines of fade-drag machinery (1,112 by the top of the stack). The fade hit-test + drag arm is a cohesive slice; lift it beside `clips_draw.rs` the way `piano_roll/loop_drag.rs` was split.

## Nits

3. **note** — the drag math converts pixels to beats to frames (`x_to_beat` then `* spb`), which can jitter ±1 frame against the direct pixel-to-frame mapping used by `fade_handle_xs`; harmless, but a single conversion helper would keep handle position and drag value exactly reciprocal.
4. **note** — no CI has run on this stacked branch; confirm checks when it retargets dev after #196 merges.

alexanderwanyoike:
Addressed in 9946a33. Unchanged rounded fade values now defer the project edit decision and produce no dirty flag, Undo snapshot, Section refresh, or engine command. Fade hit-testing and reciprocal pixel/frame drag geometry moved into dedicated timeline modules, bringing clips.rs below 1,000 lines. Focused tests and strict Clippy pass.

alexanderwanyoike:

Both findings resolved in 9946a33 and the stack:

1. No-op suppression — **resolved better than asked**: the drag handler equality-early-returns, dirty flows through `action.mark_dirty`, and the growing snapshot/dirty special-cases were consolidated into one `defers_project_edit()` predicate used by both the app update path and the section timeline path, with a regression test. On the #198 tip the audible-length comparison runs *before* the crossfade unlink, exactly as specified, so a zero-movement drag neither dirties nor dissolves a link.
2. `clips.rs` file size — **resolved**: split into `clip_drag.rs` and `fade_drag.rs`; `clips.rs` now sits at 999 lines. One standing caution shared with #198: two files in this stack now sit at exactly 999, so the very next edit to either must split first per the split-before-crossing rule.

PR #198, opened 2026-08-24
feat: add linked audio clip crossfades

alexanderwanyoike:
## Summary

- create an explicit equal-power crossfade from two selected Audio Clips with one shared-edge overlap
- persist reciprocal Clip links and repair stale or incomplete relationships during project load
- render complementary curves through the shared live, Section, Bounce, and export path
- show curved envelopes and crossfade marks in Arrange and Section timelines
- let a middle Clip keep independent incoming and outgoing crossfades
- unlink only an edited fade edge, or both edges for geometry changes, while retaining ordinary fade lengths

## Stack

- depends on #196 and #197

## Validation

- `cargo test --workspace`

### DIFF
--- a/crates/vibez-core/src/track.rs
+++ b/crates/vibez-core/src/track.rs
@@ -334,7 +334,7 @@ impl ClipFades {
         }
     }
 
-    /// Per-frame linear amplitude. This is safe to call on the audio thread.
+    /// Per-frame amplitude. This is safe to call on the audio thread.
     #[inline]
     pub fn gain_at(self, clip_frame: u64, duration: u64) -> f32 {
         if clip_frame >= duration {
--- a/crates/vibez-core/src/track.rs
+++ b/crates/vibez-core/src/track.rs
@@ -146,6 +146,10 @@ pub struct ClipFades {
     fade_in_frames: u64,
     #[serde(default)]
     fade_out_frames: u64,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    crossfade_in_from: Option<ClipId>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    crossfade_out_to: Option<ClipId>,
 }
 
 impl<'de> Deserialize<'de> for ClipFades {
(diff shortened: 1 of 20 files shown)


alexanderwanyoike:
# mini-code-review — verdict: merge as-is (after #197's fix lands beneath it)

The crossfade engineering is careful and I could not break it. The equal-power math is exact: the outgoing curve is `sin((frames_after+1)/N * pi/2)` and the incoming `sin(k/N * pi/2)`, which are sin/cos complements over the overlap, so `gain_out^2 + gain_in^2 == 1` at every frame — verified analytically and pinned by both the value-object test and the through-the-engine render test. Link integrity is defended everywhere I probed: every geometry edit (move, resize, source bounds, warp, clear-warp, quantize, join, split, delete, cross-track move) unlinks first while retaining the audible fade lengths; copy/paste/duplicate strip links so a pasted clip cannot point at a peer it no longer overlaps; `for_fragment` and the split path null links on fragments and repair the peer; and `sanitize_loaded_crossfades` validates reciprocity plus geometry (strict ordering, shared-edge overlap, matching fade lengths) on load with a repair test. The middle-clip case (independent in and out links) is tested. `sin()` per frame in the callback is allocation-free and cheap enough for the RT budget.

## Design / DRY

1. **fix (next PR in this area, per the split-before-crossing rule)** — `crates/vibez-ui/src/domains/arrangement/media_ops.rs` lands at 999 lines, one line under the hard 1,000 limit. Whoever touches it next must split first; flagging now so it does not become a surprise blocker mid-feature.

## Nits

2. **note** — `crates/vibez-ui/src/widgets/timeline/clips_draw.rs:1295` — the crossfade mark is a literal `"X"` text glyph among canvas drawing; the same letter-as-icon pattern the Browser-icons review (#186) replaced with Lucide codepoints. Canvas `fill_text` accepts a font, so an icon glyph (or a small drawn mark) would match the app's icon language.
3. **note** — `crates/vibez-ui/src/app/project_sections.rs:353` — `valid_crossfade_pair` takes eight positional arguments behind a `clippy::too_many_arguments` allow; two small `{id, start, duration, fades}` structs would remove the allow and the call-site argument-ordering risk.
4. **note** — no CI has run on this stacked branch; confirm checks when the stack retargets dev.

alexanderwanyoike:
Follow-up polish landed in 1c44f3c after rebasing onto #197. A zero-movement fade drag now preserves linked crossfades, the canvas draws the crossfade mark directly instead of using a text glyph, and project-load validation uses a named candidate type instead of eight positional arguments. Focused tests and strict Clippy pass.

alexanderwanyoike:

All notes resolved in 1c44f3c7:

1. The `"X"` text glyph is now a drawn cross path with the accent stroke, matching the canvas idiom.
2. `valid_crossfade_pair` takes two `LoadedCrossfadeCandidate` structs — the clippy allow is gone and the location/track equality checks folded into the predicate, which is stronger than the original eight-argument version.
3. `media_ops.rs` and `clips.rs` both sit at exactly 999 lines. Flagging once more: the split-before-crossing rule binds whoever touches either file next; trimming to one line under the ceiling defers the split, it does not satisfy it.

PR #199, opened 2026-08-24
feat: add nondestructive audio clip reverse

alexanderwanyoike:
## What changed

- adds a persisted, typed Forward or Reverse playback direction to Audio Clips
- adds a clear REVERSE ON/OFF control in Clip View and Toggle Reverse in the clip menu
- reverses the resolved visible Clip sequence, including looped Clips and Clips with a distinct Start marker
- keeps fades and crossfades attached to timeline edges
- mirrors the detailed waveform, source markers, loop brace, Start marker, and timeline waveform preview
- preserves reverse playback through copy, duplicate, move, split, Track Mute trim, Sections, project save/reopen, and offline bounce
- keeps source audio immutable and updates the engine in place
- moves fragment geometry out of media_ops.rs so that file remains below the 1,000-line ceiling

## Review notes

Playback direction lives directly on ClipInfo, UiClip, and EngineClip. There is no parallel ID collection to keep in sync. Reverse maps clip-local time before the existing FrameClipTimeline resolves Start and loop geometry.

Split and trim calculate phase from the opposite Clip edge while reversed. This preserves the exact audible sequence across both ordinary and looped fragments.

Reverse deliberately keeps existing crossfade links because Clip geometry does not change. Reversing either Clip may change the character of the sound entering the overlap, while the linked equal-power fade remains attached to the same timeline edges.

## Verification

- cargo fmt --all
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --workspace --all-features
- 1,168 workspace tests passed, 2 doc tests ignored


alexanderwanyoike:

This is how the `ClipTimeline` investment was supposed to pay off, and it does. Reverse is one composition point: `map_clip_frame` mirrors clip-local time *before* the timeline resolves Start and loop geometry, so pickups, loop wraps, and the start marker all reverse correctly with no new mapping code; fades key off the unreversed frame so they stay glued to timeline edges (both behaviours pinned by engine tests, including the resolved-loop traversal). The split/trim math measures fragment phase from the opposite edge under reverse (`audio_fragment_source_start`), fragments inherit the direction so the audible sequence is exactly preserved, and the detail widget mirrors peaks, markers, brace, and drag mapping through one shared fraction flip — while also replacing its last hand-inlined loop-modulo with `timeline.source_at`, a further consolidation. Source audio stays immutable; the engine updates direction in place. Persistence uses the established `skip_serializing_if` + legacy-default pattern with tests. The promised `media_ops.rs` relief landed as the extracted `fragment_geometry.rs`.

## Nits

1. **note** — toggling Reverse intentionally keeps crossfade links (geometry is unchanged even though the audible content entering the overlap changes completely). Defensible, but it is the one place a linked equal-power pair can change character without an unlink; worth a line in the PR body.
2. **note** — no CI on the stacked branch; confirm checks as the stack retargets dev.

alexanderwanyoike:
Captured the crossfade behavior in the PR body. Reverse intentionally preserves linked equal-power fades because timeline geometry is unchanged, while noting that the audible material entering the overlap can change. The full rebased stack also passes workspace tests and strict Clippy locally.

PR #200, opened 2026-08-24
feat: add editable audio transient markers

alexanderwanyoike:
## Summary

- persist ordered source-frame transient markers without changing clip playback or source media
- detect onset suggestions off the UI and audio threads, while preserving markers added or moved by the producer
- draw, select, add, drag and delete markers in the shared Arrange and Section Audio Clip editor
- preserve marker meaning through Reverse, trim, copy, split, Warp, Transpose, Capture, Undo and save/reopen

## Stack

- targets `feat/audio-clip-reverse` from #199
- implements Production Card 16
- Warp anchors and slicing remain separate follow-up PRs

## Verification

- `cargo test --workspace --all-features`
- 630 `vibez-ui` tests passed, full workspace passed, 2 doc tests ignored by design


alexanderwanyoike:

The marker model is right: `TransientMarkers` canonicalizes on every construction and deserialization (sorted, deduped, authored-wins tie-break at equal frames), markers anchor to *source* frames so Reverse, looping, and the Start marker compose for free, moving a suggestion converts it to authored per the roadmap contract, and detection runs off-thread with the established Arc-identity staleness guard plus a visible "ignored after Clip audio changed" status. Preservation is systematic: warp/transpose renders scale by the buffer ratio, splits/loads/trims retain the visible source range on every fragment path I checked (both split halves, clipboard fragments, load clamp). Widget interaction (select, add, drag-author, delete) is tested including the drag-authors-a-new-position case.

## Design / DRY

1. **fix** — `crates/vibez-ui/src/app/transient_markers.rs:56` (`apply_detected_transients_at`) — this is the third copy of the three-way timeline-location dispatch (Arrange / selected Section with commit+refresh / background Section with editor rebuild); `media_import.rs` carries the same match for quantize and BPM-detect results. Each copy re-states the selected-section commit and refresh choreography, which is exactly the kind of invariant that drifts. Extract one `with_timeline_editor_at(location, |editor, engine| ...)` helper in the app layer and route all three result appliers through it.

## Nits

2. **note** — `retain_source_range` keeps a marker exactly at `source_end` (inclusive upper bound) while the playable range is exclusive; harmless for display, just a boundary convention to keep in mind when slicing consumes markers.
3. **note** — no CI on the stacked branch; confirm checks as the stack retargets dev.

alexanderwanyoike:
Addressed in 1841e17.

- Added one shared `with_timeline_editor_at` route for Arrange, the selected Section and background Sections.
- Moved transient detection, Audio Quantize, BPM detection and Clip Warp result application onto that route.
- Section commit and refresh now live in one place and refresh only when the action is dirty.
- `cargo test -p vibez-ui --lib` passes with 630 tests.
- Strict workspace Clippy passes.

alexanderwanyoike:

Finding 1 is resolved better than asked in 1841e17: `with_timeline_editor_at` in the new `app/timeline_results.rs` now owns the three-way location dispatch, with a typed `TimelineResultEngine` (Arrange sends to the queue, Section discards) replacing the ad-hoc `DiscardingEngine` juggling, and all three result appliers (quantize, BPM-detect, transient detection) route through it — net 87 lines removed. One deliberate behaviour tightening verified as correct: the section refresh is now gated on `action.mark_dirty`, where the old quantize arms refreshed unconditionally. Notes 2 and 3 stand as informational.

PR #201, opened 2026-08-24
feat: add audio warp markers

alexanderwanyoike:
## Summary

- add persisted, monotonic Warp Markers with fixed Clip boundary anchors
- edit marker timing directly in the Audio Clip waveform with snapping and Undo
- use the same piecewise map for realtime playback, Sections, Bounce and offline rendering
- preserve maps through reverse, looping, resize, copy, split, trim, capture and save/reopen
- keep Clear Warp and whole-Clip Warp behavior deterministic

## Warp Marker playback

Warp Marker V1 is varispeed. Moving an anchor changes timing and pitch within that segment. The Audio Clip editor states this directly and the decision is recorded in Production Card 17. Pitch-preserving segment rendering is a known follow-up. Existing whole-Clip WARP remains pitch-preserving.

## Validation

- `cargo test --workspace --all-features`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `git diff --check`
- 635 `vibez-ui` tests passed, full workspace passed, 2 doc tests ignored by design

Stacked on #200.

PR #202, opened 2026-08-24
feat: slice audio clips at markers

alexanderwanyoike:
## Summary

- add explicit Slice at Transients and Slice at Warp actions to the Audio Clip editor
- convert every audible marker occurrence through loop, reverse and piecewise-warp timing
- split looped material at wrap points and create independent one-shot Clips without stale loop geometry
- preserve unrelated timeline and piano-roll selection while selecting the new Clips
- share one fragment constructor across split, slice and Track Mute trim operations
- reject operations above 256 slices rather than flooding the engine command queue
- split the arrangement domain tests into focused modules below the 1,000-line file limit
- make the complete multi-slice operation one Undo step

## Validation

- `cargo test -p vibez-ui --lib` (643 passed)
- `git diff --check`

Stacked on #201.


alexanderwanyoike:
# mini-code-review — verdict: merge after fixing findings 1, 2, 4, 5, 7

The slice-point math is right where it matters most: boundaries are exact by construction (sorted, deduped, gap-free, durations sum to the original), transient source frames route through the warp map and `occurrences_of` so looped clips cut at every audible occurrence, fragments re-derive their piecewise maps, `Arc<DecodedAudio>` sharing is pinned by `Arc::ptr_eq`, the whole slice is one undo step through `defers_project_edit`, and sample-exact playback preservation was verified empirically for the looped case the tests do not assert.

## Bugs

1. **fix (small)** — `crates/vibez-ui/src/domains/arrangement/slicing.rs:291` — reversed clips cut on the wrong side of the marker by one frame. `occurrences_of` yields play position `p`; the audible frame under reverse is `duration - 1 - p` (`map_clip_frame`), but the cut lands at `duration - p`, so the transient sample ends the *previous* slice instead of starting the next one, unlike the forward case which is exact. Verified empirically (marker at source 300 on a 1,000-frame reversed clip: slices `[0..700)` ending on source 300, `[700..1000)` starting on 299). Inaudible in isolation, but it is an inconsistent convention that the new reversed-warp test pins in place, making it harder to change later — and #203 inherits it for pad boundaries.

2. **fix** — `slicing.rs:65-68` — slicing an audio clip blanks the piano roll and wipes unrelated selection. `self.selected_note_clip = None` closes the MIDI editor the user had open on a different track (`views_detail.rs:53` reads it), and `selected_clips.clear()` drops selection on every other track; `op_split_audio_clip` does neither. Replace the original's selection entry with the slices and leave the note-clip selection alone.

3. **note (pre-existing, this stack)** — slices of a looped clip inherit `loop_enabled` plus the original loop region while `source_offset` becomes the wrapped phase, so a slice can hold `loop_start < source_offset`; the first resize then runs `clamp_loop_to_clip` and silently changes what the slice plays. Inherited from the #201 fragment geometry, but "slice a loop into hits, then nudge one" is this feature's headline workflow — worth fixing somewhere in this stack (disable looping on slices, or re-derive the region per fragment).

## Design / DRY

4. **fix** — `media_ops.rs:686` and `slicing.rs:222` — both crossfade-unlink additions are dead code. `ClipFades::for_fragment` delegates to `ClipFades::new`, which hardcodes both links to `None`, and every `replace_audio_clip` caller builds fragment fades through `for_fragment`; deleting both additions (verified empirically) changes nothing and the suite still passes. Either drop them, or keep only the `replace_audio_clip` chokepoint loop with a comment saying it defends the invariant — and reword the PR bullet, which currently reads as a bug fix for a bug that does not exist.

5. **fix** — fragment construction is now triplicated (`slicing.rs:213`, split L/R at `media_ops.rs:777`, trim-by-mutes at `media_ops.rs:89`): the same seven steps (clone, remint id, name, position, duration, `audio_fragment_geometry`, `for_fragment`, `retain_source_range`), already drifting (slicing adds `.unlinked()`, the others do not; field-assignment order differs). Extract `audio_fragment(original, name, local_start, duration) -> UiClip` beside `audio_fragment_geometry` and put all three callers on it.

6. **note** — `slicing.rs:69-71` — the explicit inspector/marker-selection clears are no-ops; the post-message orphan sweep (`discard_orphaned_audio_clip_inspector_edits`, `state/mod.rs:386`) already covers vanished clips, and split relies on it.

7. **fix** — no ceiling on slice count: one slice per marker occurrence times loop repetitions means a dense-marker looped clip can yield tens of thousands of clips and `AddClip` commands from one button press, stalling the UI with no confirmation. Add a sanity cap with a status message (mirror #203's cap, but say when it truncates).

## Nits

8. **note** — `"{} Slice {}"` nests on re-slice ("Kick Slice 2 Slice 1"); same shape as split's L/R naming, so possibly deliberate.
9. **note** — test gaps worth closing cheaply: fade edge-ownership across slices (only first keeps fade-in, only last keeps fade-out) is the semantics most likely to regress and is unasserted; the loop test asserts durations but not playback equivalence (verified to hold, so the assertion is free); no reversed-and-looped case.
10. **note** — `domains/arrangement/tests.rs` is at 3,001 lines (+142 here). If the 1,000-line rule covers test files it is the largest violation in the repo; worth an explicit ruling and, either way, a split by feature area.

alexanderwanyoike:
Addressed the full review in 47202d1.

- corrected reversed cut ownership and added the concrete source-frame regression
- preserved unrelated Clip selection and the visible MIDI editor
- made looped slices independent one-shots, with loop wraps promoted to slice boundaries so playback stays intact
- retained only the defensive crossfade unlink at the replacement chokepoint and corrected the PR description
- extracted one shared Audio Clip fragment constructor for split, slice and Track Mute trim
- removed redundant inspector and marker cleanup
- added a 256-region rejection limit with no partial mutation
- added fade ownership, playback equivalence, reversed-loop and limit coverage
- split the 3,001-line arrangement test file into nine focused modules; the largest is 736 lines

Validation is green with all 643 `vibez-ui` library tests and strict Clippy.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge as-is (in stack order; CI on retarget)

All findings resolved in 7c70535 + d5b0223, several beyond the ask:

1. Reverse one-frame cut — **resolved**: marker cuts use `duration - 1 - position` so the audible transient starts the next slice; pass-boundary wrap cuts correctly keep `duration - position` (boundaries are between frames, markers are on them — the distinction is right).
2. Piano-roll blanking / selection wipe — **resolved**: only the sliced clip's own selection entry is removed; note-clip and marker selections are untouched.
3. Loop-slice stale bounds — **resolved beyond the ask**: slices disable looping and `slice_boundaries` adds every loop-wrap position as a mandatory boundary, so each slice is a true one-shot over a contiguous source range with no stale loop geometry. This also fixes the wrap-straddling problem at the shared level for #203.
4. Dead crossfade-unlink code — **resolved**: removed, with the chokepoint loop kept and honestly commented as defensive.
5. Fragment-construction triplication — **resolved**: one `audio_fragment` constructor in `fragment_geometry.rs`; all three callers (slice, split L/R, trim-by-mutes) converted.
6. Redundant cleanup — **resolved** (dropped alongside 2).
7. Slice-count ceiling — **resolved**: `MAX_TIMELINE_SLICES = 256`, refusing with an explicit count message rather than truncating silently — better than the truncate-and-say approach I suggested.
8-10. Test gaps closed (edge-ownership assertions present), tests split into `tests/` modules per the file-size ruling, and context-menu slice entries added for waveform-context parity.

PR #203, opened 2026-08-24
feat: slice audio clips to drum rack

alexanderwanyoike:
## Summary

- add Slice to Drum Rack from the Audio Clip editor
- prefer Transient Markers and fall back to Warp Markers
- flatten the Clip audio heard through loop, Reverse, Warp, Transpose, gain and edge fades off the UI thread
- stage that flattened audio as project-owned media before creating the rack, so in-session and reopened playback use the same bytes
- create a native Drum Rack Project Track with up to sixteen pad zones and an explicit truncation status
- generate a full-velocity MIDI Clip that reconstructs the slice timing without hidden attenuation
- preserve pad bounds, names, gain, Project Media and MIDI notes through save/reopen
- share the canonical slice-boundary walk, Drum Rack layout constants and numbered Project Track factory
- seed the complete new Track through canonical project replay in Arrange and Section construction
- make Track, instrument, pads and MIDI Clip one Undoable project edit

## Validation

- `cargo test --workspace --all-features` (651 UI tests; full workspace green)
- `git diff --check`

Stacked on #202.

### DIFF
--- a/crates/vibez-core/src/track.rs
+++ b/crates/vibez-core/src/track.rs
@@ -482,8 +482,29 @@ impl MediaSourceRef {
 }
 
 /// Persisted state for a future native drum rack.
+pub const DRUM_RACK_PAD_COUNT: usize = 16;
+pub const DRUM_RACK_BASE_NOTE: u8 = 36;
+
+pub const fn drum_rack_pad_pitch(pad_index: usize) -> Option<u8> {
+    if pad_index < DRUM_RACK_PAD_COUNT {
+        Some(DRUM_RACK_BASE_NOTE + pad_index as u8)
+    } else {
+        None
+    }
+}
+
+pub const fn drum_rack_pad_index(pitch: u8) -> Option<usize> {
+    if pitch >= DRUM_RACK_BASE_NOTE && pitch < DRUM_RACK_BASE_NOTE + DRUM_RACK_PAD_COUNT as u8 {
+        Some((pitch - DRUM_RACK_BASE_NOTE) as usize)
+    } else {
+        None
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct DrumPadState {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
     pub source: Option<MediaSourceRef>,
     pub gain: f32,
     pub pan: f32,
--- a/crates/vibez-instruments/src/drum_rack.rs
+++ b/crates/vibez-instruments/src/drum_rack.rs
@@ -3,12 +3,10 @@ use std::sync::Arc;
 use vibez_core::audio_buffer::DecodedAudio;
 use vibez_core::effect::ParamDescriptor;
 use vibez_core::midi::InstrumentKind;
-use vibez_core::track::DrumPadState;
+use vibez_core::track::{drum_rack_pad_index, DrumPadState, DRUM_RACK_PAD_COUNT};
 
 use crate::Instrument;
 
-const PAD_COUNT: usize = 16;
-const BASE_PAD_NOTE: u8 = 36;
 const MAX_VOICES: usize = 32;
 
 #[derive(Debug, Clone)]
@@ -91,14 +89,13 @@ impl DrumRack {
     pub fn new(sample_rate: f32) -> Self {
         Self {
             sample_rate,
-            pads: (0..PAD_COUNT).map(|_| Pad::default()).collect(),
+            pads: (0..DRUM_RACK_PAD_COUNT).map(|_| Pad::default()).collect(),
             voices: (0..MAX_VOICES).map(|_| Voice::inactive()).collect(),
         }
     }
 
     fn pitch_to_pad(pitch: u8) -> Option<usize> {
-        let offset = pitch.checked_sub(BASE_PAD_NOTE)? as usize;
-        (offset < PAD_COUNT).then_some(offset)
+        drum_rack_pad_index(pitch)
     }
 
     fn frame_range(pad: &Pad, sample: &DecodedAudio) -> Option<(usize, usize)> {
@@ -301,9 +298,9 @@ impl Instrument for DrumRack {
             pad.fine_tune = state.fine_tune;
             pad.one_shot = state.one_shot;
             pad.choke_group = state.choke_group;
-            if pad.sample_name.is_none() {
-                pad.sample_name = state.source.as_ref().map(|source| source.display_name());
-            }
+            pad.sample_name = state
+                .name
+                .or_else(|| state.source.as_ref().map(|source| source.display_name()));
         }
     }
 }
--- a/crates/vibez-project/tests/project_format_v1.rs
+++ b/crates/vibez-project/tests/project_format_v1.rs
@@ -286,7 +286,8 @@ fn sliced_drum_rack_pads_and_reconstruction_notes_survive_reopen_with_one_media_
     let source = || MediaSourceRef::LocalFile {
         path: source_path.clone(),
     };
-    let pad = |start, end| DrumPadState {
+    let pad = |name: &str, start, end| DrumPadState {
+        name: Some(name.into()),
         source: Some(source()),
         gain: 1.0,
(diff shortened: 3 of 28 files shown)


alexanderwanyoike:

The scaffolding is genuinely good: the transaction is atomic (snapshot before routing, one `push_edit`, all four failure paths bail before any mutation, undo removes track + instrument + MIDI clip together, all tested), the engine is seeded through the canonical `replay_track_to_engine` path with the section-refresh reordering handled correctly, memory is 16 Arc bumps with fractional pad windows (pinned by `Arc::ptr_eq` and the one-media-row persistence test), and note timing matches #202's audible slice positions in the ordinary case. But the core slice-window derivation is wrong for looped sources — which is the primary material this feature exists for — plus reconstruction fidelity gaps that the PR body currently overclaims.

## Bugs

1. **fix (blocking)** — `crates/vibez-ui/src/domains/arrangement/slice_to_drum_rack.rs:66` — looped source clips produce wrong or identical pads. The code binds element 0 of `audio_fragment_geometry` as the pad start, but for a looping clip that branch returns `(clip.source_offset, timeline_start, map)` (`ui_types.rs` `warp_geometry_for_fragment`): element 1 carries the fragment's real phase; #202's slicing binds all three. Verified against source and empirically: a warped looped clip yields every pad as `(0.0, 1.0)` (all pads play the whole source; the MIDI clip retriggers the full loop per slice), and an unwarped looped clip's wrap-straddling slice comes out `(0.25, 0.75)` while the clip audibly plays 250..500 then wraps to 0..250. Minimum fix: use element 1 as the pad start; the wrap-straddling slice cannot be one contiguous pad range at all, so guard it (status) or flatten the loop first.

2. **fix (guard or document)** — reversed clips are not reconstructed: pads have no direction (`DrumPadState` has no reverse field; the rack only advances forward), so slicing a reversed clip yields the correct regions in the correct order, each playing *forwards* — audibly different from the source. Either block with a status for reversed clips or state the limitation in the PR body and UI.

3. **fix** — `slice_to_drum_rack.rs:21` — `SLICE_VELOCITY = 100` makes every slice play 2.1 dB below the source (the rack scales by `velocity/127`); use 127 for faithful reconstruction, and reword the body's "velocity-aware MIDI Clip", which is a constant. (The additional pad-pan attenuation the probe measured is the rack's inherent center pan law affecting all pads, not this PR's to fix.)

4. **fix (decision + documentation)** — warped or transposed clips diverge after save/reopen: `pad.audio` is the in-session stretched/transposed buffer but `pad.source` is the raw file, and reopen reloads the raw file with `coarse_tune`/`fine_tune` hardcoded 0 — so a break warped from 120 to 140 BPM reconstructs on-grid until the project is reopened, then every pad reverts to source tempo and pitch. The body's "preserve pad bounds through save/reopen" is true of bounds only. Either persist the rendered slices as project media, compensate via tune fields, or guard warped/transposed sources with a status; silently drifting after reopen is the one wrong option.

5. **fix (small)** — `slice_to_drum_rack.rs:60` — the 16-slice cap silently absorbs everything past cut 15 into the last pad and the status still reads "16 slices" with no mention of dropped markers. Say when it truncates.

## Design / DRY

6. **fix** — `MAX_SLICES`/`FIRST_PAD_NOTE` restate the rack's private `PAD_COUNT`/`BASE_PAD_NOTE` (`drum_rack.rs:10`), and `36 + index` is already spelled out in `views_devices.rs` and pad init sites. Publish the constants or a `drum_pad_pitch(index)` helper so the rack cannot be resized or rebased without every consumer following.

7. **fix** — fourth copy of the track-creation sequence (unique number, counter bump, colour, id, constructor, `attach_channel_eq`, push) across `project_tracks.rs` twice, `media.rs`, and here. Extract one constructor on `ProjectTracksState`.

8. **fix** — the slice-boundary walk duplicates #202's (`boundaries` vector, `windows(2)`, `audio_fragment_geometry`, `"{} Slice {}"` naming) with a *different* tuple binding — which is exactly how bug 1 got in. A shared `slice_boundaries(clip, cuts)` used by both makes the divergence unrepresentable.

## Nits

9. **note** — pad boundaries cut hard with no fade while #202's timeline slices carry `for_fragment` fades, so rack slices will click on non-zero-crossing cuts; pads have no fade parameter, so this is a follow-up card, not a fix here.
10. **note** — `choke_group: Some(1)` on all pads makes the rack monophonic for manual playing; defensible for reconstruction, but state it.
11. **note** — pad gain clamps `gain_db.linear()` to 2.0, silently capping clips hotter than +6 dB.
12. **note** — pad names ("… Slice 1") do not survive reopen: `DrumPadState` has no name field, so pads come back named after the source file.
13. **note** — no test covers the Section-mode path, which is the one thing this PR newly coordinates across `actions.rs` and `update_timeline.rs`; a small test would pin the refresh ordering.
14. **note** — `tests.rs` 3,001 → 3,152 and `undo_tests.rs` 865 → 947; same test-file size ruling as requested on #202. The PR body bullet about splitting the arrangement dispatcher belongs to the base, not this PR (`mod.rs` goes 956 → 960 here).

alexanderwanyoike:
Addressed the complete review in c874b9d, rebased onto the corrected #202.

- fixed looped and wrap-straddling sources by reusing #202 boundaries and flattening the audible Clip timeline first
- Reverse, Warp, Transpose, Clip fades and Project Media now reconstruct after reopen from the staged heard audio
- changed reconstruction notes to velocity 127
- report when later regions are folded into pad 16
- centralized Drum Rack pad count and MIDI note mapping
- centralized numbered Project Track creation and the standard channel strip
- removed the +6 dB pad-gain clamp; Clip gain is retained as full-range pad gain
- made generated pads polyphonic instead of forcing choke group 1
- persisted authored pad names with backward-compatible serde defaults
- added looped pad-range, warped/reversed save-reopen, hot gain, name persistence and Section refresh-routing coverage
- kept arrangement and undo test modules under the 1,000-line limit

Per-pad micro-fades remain a named follow-up in Card 19 and the deeper Drum Rack milestone because the current pad state has no envelope contract yet.

Validation is green across the full workspace and strict Clippy.

alexanderwanyoike:
# mini-code-review (re-review) — verdict: merge as-is (in stack order; CI on retarget)

The needs-another-pass verdict is cleared by a redesign that removes the whole bug class rather than patching instances. The slicer now consumes a pre-rendered flattened buffer: `prepare_drum_rack_audio_async` renders the clip's audible timeline offline (through the full source mapping — loop wraps, warp map, reverse, fades — via `source_frame_position_at` + `gain_at`), encodes it to WAV, stages it as provenance-aware Project Media, and re-decodes for verification. Pad windows become simple contiguous ranges over that buffer.

1. Looped sources produced wrong/identical pads — **resolved structurally**: loops are baked into the flattened render; pad ranges are plain `local_start..local_start+duration`; `slice_boundaries` (shared with #202) adds wrap boundaries.
2. Reversed clips not reconstructed — **resolved structurally**: reversal is baked into the render, so pads reproduce the reversed audio faithfully.
3. Velocity — **resolved**: `SLICE_VELOCITY = 127`.
4. Save/reopen divergence for warped/transposed sources — **resolved**: pads reference the staged rendered media, not the raw file, so reopen reproduces exactly what was sliced.
5. Silent 16-cap — **resolved**: the confirmation dialog validates the region count against `DRUM_RACK_PAD_COUNT` and refuses with a count message; a marker-type picker makes the choice explicit.
6. Rack constants — **resolved**: `DRUM_RACK_PAD_COUNT` / `drum_rack_pad_pitch(index)` published from vibez-core and consumed here.
7. Track-creation duplication — **resolved**: `project_tracks.add_numbered_track(...)` extracted and used.
8. Boundary-walk duplication — **resolved**: shared `slice_boundaries`.
9-13. Pad fades arrive in stacked #213 (the follow-up card, as filed); choke default is now `None`; gain no longer clamped; the async preparation carries an `expected_clip` and the arrival handler guards with `has_same_audible_geometry` plus a location check, with a clear cancellation status — the right staleness discipline, and semantically sharper than Arc identity for this case.

Remaining note: pad names still do not survive reopen (`DrumPadState` has no name field) — cosmetic, unchanged from the first review.

PR #204, opened 2026-08-26
fix: preserve workspace above the detail panel

alexanderwanyoike:
## What changed
- cap the lower detail panel by both available workspace height and window proportion
- keep at least 480 px for shell chrome and the active workspace
- preserve compact Audio Clip and MIDI editor minimums when space allows

## Why
Dragging the Clip or Devices panel upward could crush Perform or Arrange into a narrow strip. The lower panel now stops before the main workspace becomes unusable.

## Validation
- cargo test -p vibez-ui app::views_detail::tests --lib
- cargo clippy -p vibez-ui --all-targets -- -D warnings


alexanderwanyoike:

Small, tested clamp preventing the detail panel from starving the workspace. The stacked #211 revises these same constants further (with a documented rationale and a regression test), so this PR is effectively the first iteration of that fix — fine to merge in stack order.

PR #205, opened 2026-08-26
fix: use a labelled reverse icon control

alexanderwanyoike:
## What changed
- replace the wide Reverse On/Off text control with the Lucide flip-horizontal icon
- retain the active accent treatment when reverse playback is enabled
- explain both states with a tooltip above the control

## Validation
- cargo check -p vibez-ui --all-targets
- cargo clippy -p vibez-ui --all-targets -- -D warnings

### DIFF
--- a/crates/vibez-ui/src/app/views_audio_clip.rs
+++ b/crates/vibez-ui/src/app/views_audio_clip.rs
@@ -1,6 +1,8 @@
 //! Audio Clip Inspector panel and waveform controls.
 
-use iced::widget::{button, canvas, column, container, horizontal_space, row, text, text_input};
+use iced::widget::{
+    button, canvas, column, container, horizontal_space, row, text, text_input, tooltip,
+};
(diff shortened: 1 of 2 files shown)


alexanderwanyoike:
# mini-code-review — verdict: merge as-is, with one drive-by fix in the same file

The labelled icon control is right: `FLIP_HORIZONTAL_2` (`E35D`) verified present in the bundled Lucide TTF cmap, active/inactive styling matches the marker-button language, and the tooltip states both states plainly.

## Bugs (drive-by, pre-existing)

1. **fix (one-liner)** — `crates/vibez-ui/src/icons.rs:37` — `SCISSORS` is `\u{E152}`, which the bundled Lucide TTF maps to the **send** (paper-plane) glyph; actual scissors is `\u{E14E}` (verified against the font's cmap). Every "Slice Clip" and "Trim Track Mutes" menu entry currently renders a paper plane. Since this PR already touches `icons.rs`, fix the codepoint here.

PR #206, opened 2026-08-26
Add editable audio fade curves

alexanderwanyoike:
Adds draggable fade-curve handles to selected audio clips. The curve is stored with the clip, survives fade-length edits and project reopen, drives both waveform rendering and real-time gain, and remains backward-compatible with existing projects. Linked crossfades keep their equal-power behaviour.

Validation
- cargo test -p vibez-core --lib
- cargo test -p vibez-ui --lib
- cargo clippy -p vibez-core -p vibez-ui --all-targets -- -D warnings
- cargo fmt --all --check


alexanderwanyoike:
# mini-code-review — verdict: merge after fixing findings 1, 2, and the file splits (6)

The curve model is a good choice: one persisted percent in [-100, 100] driving a power curve with a closed-form inverse for the drag. Conventions hold: no-op suppression precedes the crossfade unlink, one UndoGestureId per drag, `SetAudioClipFadeCurve` registered in `defers_project_edit`, engine re-clamps, offline render parity via `gain_at`.

## Bugs

1. **fix (one-liner)** — `crates/vibez-core/src/track.rs:184` — `FadeCurve::gain` computes two `powf` per call with no shortcut for the linear default, and `gain_at` runs once per output frame per clip in the realtime callback. Every existing project's fades regress from one division to two transcendentals per frame, and the exponent (a per-clip constant) is recomputed 48,000 times a second. Add `if self.0 == 0 { return progress.clamp(0.0, 1.0); }` at the top.

2. **fix** — `crates/vibez-ui/src/widgets/timeline/clips.rs:462` — `fade_curve_hit` is tested before `fade_handle_hit`, and at steep curves the curve handle's y enters the fade-length handle's band (y in [21, 35] at the 70 px row height); horizontally they are half the fade width apart, so any fade narrower than ~14 px collides. A clip with a steep short fade-in can no longer have its length dragged — every press grabs the curve. Order by proximity or separate the bands.

## Design / DRY

3. **fix** — the gain law is written twice: `ClipFades::gain_at` for audio and a re-implemented closure in `clips_draw.rs:271` for the drawn envelope — the one representation the user edits against is the one that can drift from what they hear. Expose `ClipFades::gain_at_progress(progress, edge)` and draw through it.

4. **fix** — `ClipFades`'s hand-rolled `Deserialize` (`track.rs:215`) is the only path that skips the "curve is linear when the fade is zero-length" invariant every constructor enforces; `{"fade_in_frames":0,"fade_in_curve":50}` loads into an unrepresentable state and flips `is_neutral`. Normalise in the deserializer.

5. **fix (small)** — `FadeCurve::deserialize` (`track.rs:146`) parses `i8` then widens for the clamp, so any persisted value outside -128..=127 hard-fails the whole project load while the clamp only covers the ±101..±127 sliver. Deserialize `i16` and clamp — that is the impl's whole purpose.

## Nits

6. **fix (file rule)** — this PR pushes `widgets/timeline/clips.rs` 999 → 1,020 and `vibez-core/src/track.rs` 973 → 1,146 over the 1,000 hard limit; the split belongs here (ClipFades + FadeCurve are a natural `track/fades.rs` lift). `domains/arrangement/mod.rs` was already over on the base — don't feed it.
7. **note** — the envelope band (top, bottom) is derived independently in three places that agree by arithmetic coincidence; one helper should own it.
8. **note** — `is_neutral`'s new curve checks become provably redundant once finding 4 lands; keep the invariant in one home.

alexanderwanyoike:
Addressed the review findings in 2c88453:

- added the zero-curve realtime fast path and shared one gain law between audio and drawing
- resolved overlapping curve/length handles by nearest drawn distance
- widened persisted curve decoding before clamping and normalised zero-length fade state on load
- split the oversized core fade model, timeline hit testing, and arrangement fade edits into focused modules
- added regressions for wide persisted values, zero-length invariants, and overlapping handles

Verified with focused core/UI tests and clippy with warnings denied. Leaving merge ownership with Claude as agreed.

alexanderwanyoike:

All findings resolved in 2c88453: the linear-path `powf` shortcut is in (`if self.0 == 0`); overlapping fade controls resolve by drawn distance with the length handle winning ties (documented why); one `gain_for` law shared by rendering and the drawn envelope; `ClipFades::deserialize` normalises zero-length fades to a linear curve; `FadeCurve::deserialize` parses `i16` and clamps; and the file rule is satisfied by real splits — `track/fades.rs` (392), `clips/hit_test.rs`, `fade_edits.rs`, leaving `track.rs` 798, `clips.rs` 953, and `arrangement/mod.rs` 971, all under the limit. A reciprocal drag-math test pins the handle/inverse agreement.

PR #207, opened 2026-08-26
Make audio crossfades direct and editable

alexanderwanyoike:
Reworks crossfades around the overlap itself. Dragging a fade handle to the neighbouring clip edge creates one reciprocal crossfade, and the overlap exposes one shared slope handle that shapes both clips together. The paired curves preserve constant power, persist through project reopen and use one Undo gesture.

The context command remains as a secondary route and is now labelled Create Crossfade.

Validation
- cargo test -p vibez-core --lib
- cargo test -p vibez-ui --lib
- cargo clippy -p vibez-core -p vibez-ui --all-targets -- -D warnings
- cargo fmt --all --check

### DIFF
--- a/crates/vibez-core/src/track.rs
+++ b/crates/vibez-core/src/track.rs
@@ -795,4 +795,50 @@ mod tests {
             assert!((power - 1.0).abs() < 1e-5, "frame {frame}: {power}");
         }
     }
+
+    #[test]
+    fn shaped_crossfade_edges_remain_a_complementary_power_pair() {
+        for curve in [FadeCurve::new(-100), FadeCurve::new(0), FadeCurve::new(100)] {
+            for step in 0..=20 {
+                let progress = step as f32 / 20.0;
+                let (outgoing, incoming) = curve.crossfade_gains(progress);
+                assert!((outgoing * outgoing + incoming * incoming - 1.0).abs() < 1e-5);
+            }
+        }
+
+        let outgoing_id = ClipId::new();
+        let incoming_id = ClipId::new();
+        let curve = FadeCurve::new(65);
+        let outgoing = ClipFades::default()
+            .linked_fade_out(100, incoming_id, 100)
+            .with_linked_fade_out_curve(curve);
+        let incoming = ClipFades::default()
+            .linked_fade_in(100, outgoing_id, 100)
+            .with_linked_fade_in_curve(curve);
+        assert_eq!(outgoing.fade_out_curve(), curve);
+        assert_eq!(incoming.fade_in_curve(), curve);
+        for frame in 0..100 {
+            let power = outgoing.gain_at(frame, 100).powi(2) + incoming.gain_at(frame, 100).powi(2);
+            assert!((power - 1.0).abs() < 1e-5, "frame {frame}: {power}");
+        }
+    }
+
+    #[test]
+    fn unlinking_a_shaped_crossfade_keeps_its_length_but_returns_to_linear() {
+        let peer = ClipId::new();
+        let linked = ClipFades::default()
+            .linked_fade_out(50, peer, 100)
+            .with_linked_fade_out_curve(FadeCurve::new(80));
+        let unlinked = linked.unlink_fade_out();
+
+        assert_eq!(unlinked.fade_out_frames(), 50);
+        assert_eq!(unlinked.fade_out_curve(), FadeCurve::default());
+        assert!(unlinked.crossfade_out_to().is_none());
+
+        let reopened: ClipFades =
+            serde_json::from_str(&serde_json::to_string(&unlinked).unwrap()).unwrap();
+        let relinked = reopened.linked_fade_out(50, peer, 100);
+        assert_eq!(relinked.fade_out_curve(), FadeCurve::new(80));
+        assert_eq!(relinked.crossfade_out_to(), Some(peer));
+    }
 }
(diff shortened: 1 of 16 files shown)


alexanderwanyoike:

The core design is right and worth keeping: `crossfade_gains` warps the *phase* (`cos φ`/`sin φ` with `φ = gain(p)·π/2`), so equal-power complementarity holds identically for every curve value (verified analytically, including that both sides see the same argument at every shared frame, and that curve 0 reproduces the legacy pair exactly; pinned by a 100-frame render test across three curves). The problems are that the headline gesture cannot actually be performed, and the unlink/relink semantics audibly betray user edits.

## Bugs

1. **fix** — `crossfades.rs:30`/`:42` — crossfade creation requires the dragged fade length to equal the clip overlap to the exact frame, but the only producer is the pixel drag (`frames_at_x`, a linear pixel→frame map with no snap), making the acceptance window ~0.002 px at typical zoom (44,100 frames across 200 px). The gesture effectively never triggers below ~1 px per frame. The test dispatches the message with the exact value and never drives the canvas, so it cannot catch this. Snap when the handle is within `FADE_HANDLE_HIT_RADIUS` of the neighbour's edge, and add a canvas-driven test.

2. **fix** — even on an exact hit, the next `CursorMoved` produces a different frame value, the candidate check fails, and the fall-through path unlinks the crossfade just created — the user must *release* on the exact frame. Latch the link for the remainder of the gesture or evaluate on release.

3. **fix** — `track.rs` `unlink_fade_in/out` keep the stored curve while `gain_at` reads it under a different law per link state: a crossfade shaped to 65 sits at 0.376 linked and jumps to 0.755 the instant any unlink path fires (clip move, fade nudge, finding 2's flicker) — roughly +6 dB on the outgoing side with a simultaneous drop on the incoming (independently re-derived and confirmed). Zero the curve on unlink the way linking zeroes it, or unify the two laws.

4. **fix** — relinking discards the shape: `link_crossfade_pair` routes through `linked_fade_*`, which reset the curve to 0. Shape a crossfade, move a clip, re-crossfade — silently back to default. Carry the previous curve across relink (`with_linked_fade_*_curve` exists; nothing calls it here).

5. **fix** — the shared crossfade handle overlaps the fade-length handle at almost any curve value (0.7 px clear at curve 0; inside the band at curve ≥ ~3.5) and is hit-tested first, so one upward nudge of the curve makes the length handle unreachable. Same class as #206 finding 2; solve them together.

6. **fix or state the decision** — the always-on per-clip crossfade marker was deleted and the replacement handle draws only when a clip of the pair is selected; an arrangement no longer shows which boundaries are crossfaded at a glance.

## Design / DRY

7. **fix** — the gain-law branch is now hand-mirrored four times across two crates (twice inline in `clips_draw.rs`, replacing #206's single closure); they agree today (verified), which is exactly why divergence will be missed. Same fix as #206 finding 3.

8. **fix** — two fade-length write paths, one behaviour: the inspector's fade commit never consults `crossfade_candidate_for_fade`, so typing the exact overlap does not create a crossfade while dragging (in principle) does. Put the crossfade decision on the domain fade-length operation, not one message arm.

9. **note** — `crossfade_curve_handles` allocates a Vec and does an O(n²) clip scan on every draw, cursor move, and press on an uncached canvas; derive lazily or cache per layout.

10. **note** — the handle takes x from the geometric overlap but y from `outgoing.fade_out_curve` while the drawn envelope uses `fade_*_frames`, and the incoming curve is never read — coincident only while lengths equal the overlap, and nothing asserts it.

## File size

`clips.rs` 1,020 → 1,040, `track.rs` 1,146 → 1,206, `arrangement/mod.rs` 1,044 → 1,077 — all over the limit; whichever of #206/#207 lands first owns the split (flagged as #206 finding 6).

alexanderwanyoike:
Addressed the full review in 3fe54ca:

- fade handles now magnetically snap across the full hit radius and stay snapped across pointer moves; a canvas-driven regression covers the gesture
- the shared domain fade operation now creates crossfades for both drag and inspector edits
- unlinking returns the edge to a linear fade without a gain-law jump, while compact persisted edge memory restores the curve when the same pair is relinked
- length, fade-curve and crossfade-curve controls resolve by nearest drawn distance
- linked relationships retain an always-visible marker; the editable handle appears on selection
- drawing and realtime rendering share the same fade-in/fade-out gain methods
- crossfade layout is derived once with an ID map instead of allocating and scanning O(n²) on draw/move/press
- shared handles are only materialised when reciprocal IDs, overlap lengths and curves agree
- inherited oversized files remain below the 1,000-line ceiling

Focused core, crossfade-domain and canvas tests pass. Clippy passes with warnings denied. Leaving merge ownership with Claude as agreed.

alexanderwanyoike:

All ten findings resolved in 3fe54ca:

1/2. The drag snaps to the neighbour edge within `FADE_HANDLE_HIT_RADIUS` (exact inverse of the drawn position, commented), an already-linked pair early-returns so continuing the drag no longer destroys the link, and a canvas-driven test (`fade_drag_snaps_to_an_overlapping_neighbour_across_pointer_moves`) drives real pointer events.
3/4. Unlink zeroes the curve (no more +6 dB law jump) while stashing `previous_crossfade_*_curve`, and relink restores the preserved shape — both directions handled.
5. Handle collisions resolve through the shared distance-based `fade_control_hit`.
6. Every linked pair keeps an always-visible cross marker; selection reveals the shared curve handle (comment states the rule).
7-10. The gain law lives solely on the value object (`gain_for`/`fade_out_gain`); the inspector fade commit routes through the shared `fade_edits` path so typing the overlap links too; crossfade pairs are precomputed once into `TimelineCrossfade` (no more per-event O(n²) scan) and the handle reads the pair's own data.

PR #209, opened 2026-08-26
Add continuous transient analysis sensitivity

alexanderwanyoike:
Replaces the three immediate waveform-menu analysis commands with one Transient analysis entry and a dedicated Vibez-styled modal.

The modal has one continuous 0 to 100% sensitivity knob, an exact percentage field, mouse-wheel adjustment and clear guidance that higher values retain more attacks. Cancel, Escape and backdrop dismissal still work. Manual markers remain untouched.

Detector corrections
- sensitivity changes aubio peak thresholds without changing the established 50 ms minimum hit spacing
- aubio HFC remains the precise primary detector
- aubio energy analysis recovers bass-heavy attacks that HFC misses
- a time-domain prominence check rejects pitched decay cycles that are not useful editing boundaries
- high sensitivity can still expose those quieter events intentionally

On the exact DD 124 29E.wav loop reported in testing, the default setting now finds the 7 clear waveform attacks instead of 12 decay-cycle markers. It returns 3 strong attacks at 0%, 7 through 60%, and progressively exposes quieter events above that.

Validation
- compared the Rust HFC output with aubio 0.4.9 on the reported loop
- cargo test --workspace
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/vibez-core/src/onset.rs
+++ b/crates/vibez-core/src/onset.rs
@@ -44,13 +44,17 @@ impl TransientSensitivity {
         self.0 as f32 / Self::MAX_PERCENT as f32
     }
 
-    fn threshold_through(self, midpoint: f32) -> f32 {
+    fn log_interpolate(self, low: f32, midpoint: f32, high: f32) -> f32 {
         let normalized = self.normalized();
         if normalized <= 0.5 {
-            0.9 * (midpoint / 0.9).powf(normalized / 0.5)
+            low * (midpoint / low).powf(normalized / 0.5)
         } else {
-            midpoint * (0.001 / midpoint).powf((normalized - 0.5) / 0.5)
+            midpoint * (high / midpoint).powf((normalized - 0.5) / 0.5)
         }
+    }
+
+    fn peak_threshold(self, midpoint: f32) -> f32 {
+        self.log_interpolate(0.9, midpoint, 0.001)
     }
 }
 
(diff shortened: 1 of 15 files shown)


alexanderwanyoike:

The modal, knob, and state plumbing are clean (no debounce needed and none missing — detection dispatches once on Confirm; the staleness pattern is respected). But the headline feature is measurably broken, and the test that would have caught it was weakened in the same PR.

## Bugs

1. **fix (blocking)** — `crates/vibez-core/src/onset.rs:47` — the sensitivity knob is anti-monotonic and zeroes detection above 0%. The threshold applies *before* the 50 ms refractory (`aubio.rs`: any accepted peak claims `last_onset`) while the prominence filter runs *after*, so loosening the threshold lets low-novelty junk consume the refractory and starve the real attack that follows — which the prominence check then rejects too. Measured sweep on six identical strong tonal hits 125 ms apart: `[2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]` across 0→100% — the exact inverse of the modal's label "Higher values keep quieter attacks and create more markers." Structural fix (spanning this PR and #212): move the prominence check into the peak-acceptance path so junk never claims the refractory; then the threshold sweep behaves as labelled.

2. **fix** — `onset.rs:581` — `sensitivity_affects_yield` was weakened from `tight <= loose` to `high >= low` over five *identical* impulses, which `0 >= 0` satisfies; the regression in finding 1 passes it. Restore an assertion over a signal with graded amplitudes that requires strictly more markers at higher sensitivity.

3. **fix (small)** — the Analyse button reads `dialog.sensitivity` without normalising the text field (`update.rs`), so typing `12x` and clicking Analyse silently uses a value the user is not looking at; Enter-to-submit normalises, the button does not.

## Design / DRY

4. **fix** — `threshold_through` and `attack_prominence_floor` are the same two-segment log interpolation with different constants; one `(low, mid, high)` helper.
5. **note** — `frames_for_ms` re-derives the ms→frames conversion already inlined in the detector config.
6. **note** — the analysis modal is never cleared on clip deletion (mirrors the existing `drum_rack_slice_dialog` gap — a pattern-level fix would cover both).

## Nits

7. **note** — `parse_transient_sensitivity` re-checks a bound `TransientSensitivity::new` already clamps.
8. **note** — wheel step is 1%/notch (100 notches end to end) and Shift fine-mode applies only to drag.

alexanderwanyoike:
Addressed the review in 0a14118:

- prominence is now evaluated inside aubio candidate acceptance before a peak can claim the refractory window
- added graded-amplitude strict-yield coverage plus a complete 0–100 monotonic sensitivity sweep
- consolidated the duplicated two-segment logarithmic interpolation and the ms-to-frame conversion
- Analyse validates and commits the visible text value; invalid text keeps the modal open instead of using hidden state
- deleting a target Audio Clip now dismisses both transient-analysis and Drum Rack slicing dialogs
- wheel adjustment is 5% normally and 1% with Shift

The full workspace test suite passes, Clippy passes with warnings denied, formatting and diff checks are clean. This also establishes the acceptance hook needed by #212. Leaving merge ownership with Claude.

alexanderwanyoike:

The structural fix is exactly what the review asked for: the genuine-attack check is now a predicate passed into the detector (`detect_onsets_where`), evaluated in `process_hop` *before* a peak claims the refractory — junk candidates can no longer starve real attacks, making sensitivity monotone by construction. Verified in source. The weakened test is replaced by a strict `high > low` assertion over graded-amplitude attacks; `log_interpolate` unifies the two interpolations; `frames_for_ms` is shared; the Analyse button commits the text field before reading it; and `dismiss_clip_dialogs_for` closes both clip dialogs on deletion — the pattern-level fix covering the pre-existing drum-rack gap too.

PR #210, opened 2026-08-26
Use Drum Rack banks for audio slicing

alexanderwanyoike:
Expands the native Drum Rack from one 16-pad bank to four banks and makes Slice to Drum Rack use the complete 64-pad capacity.

What changed
- the core pad map and DSP instrument now own 64 pads across four 16-pad banks
- Slice to Drum Rack accepts up to 64 regions and maps reconstruction notes across every bank
- the slice modal reports how many banks the result will use
- Perform keeps ownership of the bracket shortcuts and moves its Instrument surface through complete 16-pad Drum Rack pages
- the Drum Rack device has separate local arrow controls with no bracket labels
- playing Perform pads in Instrument mode sends the notes for the visible Drum Rack bank
- synths and samplers retain octave navigation
- legacy 16-pad projects reopen into bank one without moving samples
- saves retain used later-bank pads while omitting empty tail banks

Root cause
The visible 4x4 surface size had been used as the Drum Rack domain capacity, so the slicer, engine and project model all inherited a UI limit.

Validation
- the Instrument path is tested through the real engine command boundary at notes 36 and 52
- cargo test --workspace
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/vibez-core/src/track.rs
+++ b/crates/vibez-core/src/track.rs
@@ -254,8 +254,10 @@ impl MediaSourceRef {
     }
 }
 
-/// Persisted state for a future native drum rack.
-pub const DRUM_RACK_PAD_COUNT: usize = 16;
+/// The rack stores four banks while the UI exposes one 4x4 bank at a time.
+pub const DRUM_RACK_BANK_SIZE: usize = 16;
+pub const DRUM_RACK_BANK_COUNT: usize = 4;
+pub const DRUM_RACK_PAD_COUNT: usize = DRUM_RACK_BANK_SIZE * DRUM_RACK_BANK_COUNT;
 pub const DRUM_RACK_BASE_NOTE: u8 = 36;
 
 pub const fn drum_rack_pad_pitch(pad_index: usize) -> Option<u8> {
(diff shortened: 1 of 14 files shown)


alexanderwanyoike:

The 4-bank expansion is cleanly built: `DRUM_RACK_PAD_COUNT` composed from bank size and count so every consumer (pitch mapping, slice validation, engine storage) follows automatically, the engine renders later banks (tested at pad 20 / note 56), and persistence migrates both directions — legacy 16-pad saves expand in place without moving pads, and saves trim trailing empty banks while always keeping at least one (both tested). The slice dialog's capacity math and the Perform surface's bank readout update coherently.

## Bugs

1. **fix (small)** — `crates/vibez-ui/src/app/views_overlays.rs:274` — the slice dialog guidance says "Switch banks with [ and ]", but no `BracketLeft`/`BracketRight` binding exists anywhere in the UI (verified by grep); the only bank controls are the ‹ › buttons on the device card. Either bind the keys or drop the sentence — shipping instructions for keys that do nothing is worse than no hint.

## Nits

2. **note** — `views_devices.rs:214` — the bank buttons use text glyphs "‹"/"›" where the icon set has `CHEVRON_LEFT`/`CHEVRON_RIGHT`; same letter-as-icon pattern previous reviews converged away from.

alexanderwanyoike:
Fixed both review points: the slicing dialog no longer advertises nonexistent bracket shortcuts, and the Drum Rack bank buttons now use the Lucide chevron icons instead of text glyphs. Focused core, instrument and UI Drum Rack tests pass; Clippy is clean with warnings denied. Leaving merge ownership with Claude.

alexanderwanyoike:

Both items resolved in cd5f09e: the phantom "[ and ]" sentence is gone from the slice dialog, and the bank buttons use `CHEVRON_LEFT`/`CHEVRON_RIGHT` icon glyphs (both verified present in the bundled Lucide TTF cmap).

PR #212, opened 2026-08-26
fix: place transient markers on audio attack boundaries

alexanderwanyoike:
## Outcome

Transient analysis now separates event detection from slice-boundary localisation. On both reported 2-second production loops it returns the seven producer-relevant attacks, removes the decay markers, and moves the retained markers back onto source-energy boundaries.

## What changed

- replace the HFC plus energy union with aubio 0.4.9 complex-domain candidates
- port adaptive whitening, log compression, phase prediction, upstream zero cases, and golden primary-event timing
- generate one fixed permissive candidate set so sensitivity only filters stable source frames
- analyse stereo channels independently to avoid polarity cancellation
- scale the FFT window at 44.1, 48, and 96 kHz
- validate candidate prominence, then backtrack the detected peak to a bounded trailing-RMS minimum
- port librosa 0.11.0 backtracking invariants with its ISC notice
- add a generated beating-tail regression fixture and timing, gain, sample-rate, and stereo-polarity coverage
- add an annotation-manifest benchmark reporting precision, recall, F1, doubled events, and boundary error percentiles

## Exact dogfood results at 50%

`LP122 01F.wav`

```text
0.247 0.482 0.737 0.981 1.228 1.466 1.718
```

`DD 124 29E.wav`

```text
0.229 0.482 0.717 0.961 1.190 1.443 1.678
```

These are seven candidates on each loop. The private 25-loop producer-approved corpus remains the release gate. Two good loops are evidence, not the whole benchmark.

## Why the diff is 782 lines

- 203 lines modify the aubio port for the complex-domain descriptor and its tests
- 210 lines are isolated generated-fixture and boundary tests
- 212 lines are the reusable private-corpus benchmark executable
- the remainder replaces the production pipeline and documents licensing and timing rules

Production `onset.rs` remains below the repository's 1,000-line ceiling after moving the new verification suite into its own module.

## Verification

- `cargo test --workspace --all-features`
- exact local analysis of both reported loops

### DIFF
--- a/crates/vibez-audio-io/examples/transient_benchmark.rs
+++ b/crates/vibez-audio-io/examples/transient_benchmark.rs
@@ -7,7 +7,11 @@ use std::{env, fs, path::PathBuf, process};
 
 use serde::Deserialize;
 use vibez_audio_io::file_io::decode_audio_file;
-use vibez_core::onset::{detect_onsets, TransientSensitivity};
+use vibez_core::onset::{
+    detect_onsets,
+    metrics::{evaluate, Evaluation},
+    TransientSensitivity,
+};
 
 #[derive(Debug, Deserialize)]
 struct Manifest {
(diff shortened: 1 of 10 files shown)


alexanderwanyoike:
# mini-code-review — verdict: merge after fixing findings 1, 2, and 3 (jointly with #209's rework)

The port work is excellent and was verified against upstream: the aubio 0.4.9 semantics (adaptive whitening, compression lambda 1.0, 4.6-hop complex delay, DC/Nyquist phase forcing, complex-domain distance) all match `onset.c`; the librosa backtracking minima rule matches `onset_backtrack` including its own test vector; licensing headers are consistent (GPL-3.0 aubio, ISC librosa notice above the derived function); nothing detection-related touches the realtime path; determinism holds; and on broadband percussion the boundary timing genuinely improves (within 2.5 ms of truth, stable, versus 209's knob-dependent drift). Saved projects are safe — markers move only on re-analysis, and the `(no change)` status avoids dirtying idempotent runs.

## Bugs

1. **fix (blocking, jointly with #209)** — sensitivity is now a complete no-op on uniform-level material: `threshold_through` and `Config::for_sensitivity` are `#[cfg(test)]` (verified), so the knob's only surviving effect is the post-hoc prominence floor. Measured: identical marker sets at 0/25/50/75/100% on uniform six-hit trains, and 2-of-6 recall on tonal clicks at every setting. This PR stabilises #209's symptom by freezing the candidate stage at one permissive threshold, but the ordering flaw (threshold before refractory, prominence after) survives. The joint fix: move the prominence check into peak acceptance so junk never consumes the refractory — then sensitivity can be rewired to the detector and behave as labelled.

2. **fix** — 3.7x slowdown plus a ~100 MB transient allocation on the *import* path: per-channel analysis doubles the FFT work and `rms_energy_envelope` allocates `vec![0.0f64; frames + 1]` (verified; 8 bytes/frame ≈ 106 MB for a 5-minute file). `detect_bpm` calls `detect_onsets` merely for a `< 8` count gate and runs on every audio import. Compute the envelope in hops (or reuse the detector's per-hop energy) and gate the count check with the cheap path.

3. **fix** — `backtrack_events_to_minima` falls back to the *raw delayed peak* (`map_or(event, ...)`, verified) when no minimum sits in the 50 ms look-back, which is `4.6 hops ≈ 26.7 ms` *after* aubio's compensated estimate — worse than not backtracking, and pinned as intended by a test. Fall back to the compensated `estimated_onset` instead, bounding the error to aubio's own accuracy.

4. **fix (small)** — the benchmark example's only `#[test]` never runs (Cargo does not run tests in example targets by default), so the scoring logic the PR body cites is unverified in CI. Finding 5's hoist fixes this for free.

## Design / DRY

5. **fix** — the greedy matcher / precision-recall-F1 scoring is duplicated wholesale between the benchmark example and `transient_analysis_tests.rs` (~60 lines each); hoist into vibez-core, making the example's test live.
6. **fix** — ~60 lines of `#[cfg(test)]`-only detector (HFC branch, `for_sensitivity`, `threshold_through`) exist solely to keep tests of dead code passing; either rewire sensitivity (finding 1) or delete the branch.
7. **note** — `canonicalize_onsets` mutates its input to sort, then returns a fresh Vec; take it by value. `backtrack_events_to_minima` linear-scans minima per event; `partition_point` is O(log n).

## Nits

8. **note** — `BACKTRACK_WINDOW_MS = 23.22` / `HOP_MS = 2.9` exist to produce 1024/128 frames at 44.1 kHz; say so.
9. **note** — `onset.rs` is at 992 lines; anything further needs the split first.
10. **note** — re-analysis moves every Suggested marker (up to 50 ms earlier), so slices or warp anchors derived from old markers will not line up; deserves a release note. Test-quality gaps worth closing: the sample-rate/gain robustness test asserts only F1 at 25 ms tolerance (no timing pin), and there is no sharp-click case — the only fixture is the material the pipeline handles best.

alexanderwanyoike:
Addressed the full review in d1f5a22:

- sensitivity now drives the complex-domain detector and producer prominence is accepted before a candidate can claim the refractory window
- removed the dead test-only HFC path and retained the aubio complex-domain upstream/golden coverage
- replaced the source-length f64 prefix allocation with a rolling window bounded by the analysis window
- BPM confidence now reuses its existing cheap onset-flux envelope instead of running the FFT transient detector again on every import
- backtracking retains aubio's compensated estimate when no valid energy minimum exists, and minima lookup uses `partition_point`
- moved localisation and benchmark metrics into focused modules; the scorer's regression now runs in normal core CI
- documented the 1,024/128-frame timing constants and added a sharp-click timing regression pinned to one 2.9 ms localisation hop
- consolidated sensitivity interpolation and moved `onset.rs` below the 1,000-line limit

Release note for the eventual release: re-analysis replaces Suggested markers using the corrected earlier attack boundaries. Authored markers remain intact, but slices or Warp Markers previously derived from old suggestions are not moved automatically.

Full workspace tests pass (including 674 UI tests), Clippy passes with warnings denied, formatting and diff checks are clean. I could not rerun the private production-loop corpus because its manifest/audio are not present in this worktree; the committed deterministic fixtures and upstream golden tests pass. Leaving merge ownership with Claude as agreed.

alexanderwanyoike:

All blocking findings resolved in d1f5a22, on top of #209's structural rework:

1. Sensitivity is rewired through the production path — the `#[cfg(test)]`-only detector branch is gone (one remaining cfg(test) in aubio.rs, previously eight), and the monotonic predicate flows into candidate acceptance.
2. The import cost is gone at both ends: `detect_bpm` no longer runs onset detection at all, and the localisation envelope is a trailing hop-based RMS with memory bounded by the analysis window (the 8-bytes-per-frame prefix array is deleted).
3. Backtracking falls back to the compensated `estimated` onset, not the delayed peak (`localize.rs:123`), bounding worst-case placement to aubio's own accuracy.
4/5. The scoring logic is hoisted into `onset/metrics.rs`, consumed by both the tests and the benchmark example — whose test therefore now runs in CI.
6-9. Dead test-only detector deleted, the localisation constants are documented in time units with the 128-sample librosa rationale, and `onset.rs` came down well under the ceiling via the `localize.rs`/`metrics.rs` splits.

PR #216, opened 2026-08-26
fix: show Drum Rack pad playback activity

alexanderwanyoike:
## What changed

- report resident MIDI clip note triggers as one bounded per-track bitmask per audio block
- illuminate the matching Drum Rack pads for Arrange, Section, live input and Note Repeat playback
- retain short flashes in view-only state so brief drum notes remain visible at the UI frame rate
- preserve activity for retry if the engine event ring is temporarily full

## Verification

- `cargo test -p vibez-engine`
- `cargo clippy -p vibez-engine -p vibez-ui --all-targets -- -D warnings`

### DIFF
--- a/crates/vibez-engine/src/engine_render.rs
+++ b/crates/vibez-engine/src/engine_render.rs
@@ -307,17 +307,15 @@ impl AudioEngine {
                 track.render(pos, frames, channels, loop_region)
             };
 
-            let triggered_notes = track.pending_note_activity();
-            if triggered_notes != 0
-                && self
-                    .event_tx
-                    .push(EngineEvent::TrackNoteActivity {
-                        track_id: track.id,
-                        triggered_notes,
-                    })
-                    .is_ok()
-            {
-                track.clear_note_activity(triggered_notes);
+            let triggered_notes = track.take_note_activity();
+            if triggered_notes != 0 {
+                // Pad flashes are cosmetic. Never retain them or let them
+                // compete with authoritative input and Capture events after a
+                // UI stall fills the shared ring.
+                let _ = self.event_tx.push(EngineEvent::TrackNoteActivity {
+                    track_id: track.id,
+                    triggered_notes,
+                });
             }
 
             if live_input.is_some_and(|input| input.target_track_raw == track.id.raw()) {
--- a/crates/vibez-engine/src/engine_render.rs
+++ b/crates/vibez-engine/src/engine_render.rs
@@ -307,6 +307,19 @@ impl AudioEngine {
                 track.render(pos, frames, channels, loop_region)
             };
 
+            let triggered_notes = track.pending_note_activity();
+            if triggered_notes != 0
+                && self
+                    .event_tx
+                    .push(EngineEvent::TrackNoteActivity {
+                        track_id: track.id,
+                        triggered_notes,
+                    })
+                    .is_ok()
+            {
+                track.clear_note_activity(triggered_notes);
+            }
+
             if live_input.is_some_and(|input| input.target_track_raw == track.id.raw()) {
                 let input = live_input.expect("checked live input");
                 if !rendered && track.instrument.is_none() {
--- a/crates/vibez-engine/src/engine_tests.rs
+++ b/crates/vibez-engine/src/engine_tests.rs
@@ -1543,6 +1543,62 @@ fn note_clip_loop_renders() {
     );
 }
 
+#[test]
+fn note_clip_reports_triggered_pitches_for_device_feedback() {
+    use vibez_core::midi::{InstrumentKind, MidiNote};
+
+    let (mut engine, mut cmd_tx, mut event_rx) = AudioEngine::new();
+    let track_id = TrackId::new();
+    let clip_id = ClipId::new();
+
+    cmd_tx
+        .push(EngineCommand::AddInstrumentTrack(
+            track_id,
+            "Drums".into(),
+            InstrumentKind::DrumRack,
+        ))
+        .unwrap();
+    cmd_tx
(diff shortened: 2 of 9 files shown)


alexanderwanyoike:
# mini-code-review — verdict: needs another pass (UI half only; the engine half is merge-ready)

The engine side is well-judged and verified clean: one `|=` per note-on beside the existing `active_notes` shifts, one guarded rtrb push per track per block scaling with note density, no allocation or locking, `size_of::<EngineEvent>()` measured unchanged at 80 bytes (existing padding absorbs the u128's alignment), and the read-push-clear ordering is single-threaded-correct with the retained mask a bounded field. The u128 covers the full pitch range so the 64-pad mapping cannot truncate. The UI half, however, defeats the feature's own purpose in two ways.

## Bugs

1. **fix** — `crates/vibez-ui/src/app/views_devices.rs:729` — only the selected bank's 16 pads can ever show activity, and the bank navigation buttons carry no activity indication (hardcoded dim styling). A kit living on pads 17-32 shows a completely dead grid during playback with nothing suggesting the action is one bank over. Light the bank buttons when off-bank pads fire.

2. **fix** — `views_devices.rs:736` — playback activity reuses the selection highlight verbatim (`active = selected || firing` drives background, text, and border together), so the selected pad can never flash and every firing pad is pixel-identical to the selected one. Select pad 1, hit play: pads 1, 3, 5 look the same and pad 1's own hits are invisible. Give activity its own visual channel distinct from selection.

## Design / DRY

3. **fix** — `engine_render.rs:310` — the retain-and-retry makes this the only event with retry semantics on a ring shared with correctness-critical events (`NoteRepeated` and `InstrumentNoteInput`, which Capture consumes, are fire-and-forget `let _ =`; verified). After a UI stall fills the 1,024-slot ring, the cosmetic backlog claims freed slots ahead of later tracks' authoritative events. Not reproduced as a dropped Capture note, but the priority inversion is structural: drop the retry, or yield when the ring lacks headroom.

4. **fix** — the event fires for every instrument track; a synth-heavy project pushes per-track-per-block events and does UI HashMap inserts for flashes nothing renders. `Instrument::instrument_kind()` makes the source-side filter one comparison, which also shrinks the ring traffic feeding finding 3. (Documented trade: a future piano-roll key flash would relax the filter.)

5. **fix (one-liner)** — `mixer.rs:376` — `clear_note_activity(delivered)` masks bits that provably cannot differ from the read in a single-threaded callback; `self.pending_note_activity = 0` says what actually happens.

6. **fix (tests)** — nothing pins the ring-full retain-and-retry (fill the ring, render, drain, render, assert the event lands on block two), nor the note-repeat and live-input flash paths the PR body leads with.

## Nits

7. **note** — `state/mod.rs:126` — the map stores an *expiry* instant but is named `drum_pad_activity`; `drum_pad_flash_until` would carry it.
8. **note** — a bit retained across a ring-full block survives a solo-out or stop and flashes late on the next render; only reachable after an exact-block ring-full.
9. **note** — `retain` never shrinks capacity and the per-frame prune runs even with no rack open; bounded and cheap, just unconditional.
10. **note** — the fixed 140 ms hold renders rolls faster than ~7 Hz as a continuously lit pad rather than retriggers; fine for V1, worth knowing.

alexanderwanyoike:
Addressed the full review in `3cc3f96`:

- off-bank activity now lights the appropriate bank-navigation direction
- selected and firing pads use distinct visual states; firing wins when both apply
- clip feedback is Drum Rack-only at the engine source
- cosmetic activity is taken once and dropped if the shared event ring is full, so it cannot retry ahead of Capture/input events or flash late
- renamed the expiry map and its API around `flash_until`
- centralized clip, Note Repeat, and live-input flash mapping

Regression coverage now includes non-Drum-Rack filtering, a genuinely full 1,024-event ring with no late replay, every feedback event path including note-off exclusion, off-bank lookup, expiry, and selected-plus-firing precedence.

Local verification:
- `cargo fmt --check`
- `cargo test -p vibez-engine` (242 passed)
- `cargo test -p vibez-ui --lib` (678 passed)
- `cargo clippy -p vibez-engine -p vibez-ui --all-targets -- -D warnings`

alexanderwanyoike:

All findings resolved in 3cc3f96, each pinned by a test:

1. Off-bank blindness — **resolved**: `drum_pad_bank_is_flashing` drives the bank chevrons, so activity outside the selected bank lights the navigation (`drum_pad_bank_activity_finds_flashes_outside_the_selected_bank`).
2. Selection/activity conflation — **resolved**: a `DrumPadVisualState { Idle, Selected, Firing }` tri-state gives firing its own brighter channel (accent fill, dark text) with precedence over selection, so the selected pad flashes too (`firing_has_precedence_over_selection`).
3. Retry priority inversion — **resolved**: the push is fire-and-forget with the constraint stated in a comment (cosmetic events never compete with authoritative Capture events), and the drop decision is pinned (`drum_pad_activity_dropped_by_a_full_event_ring_is_not_replayed_late`). The leftover-bit edge cases dissolve with the retry.
4. All-instruments emission — **resolved at the source**: bits are set only when `instrument_kind() == DrumRack` (`synth_note_clip_does_not_report_drum_pad_activity`).
5. Dead-general clear — **resolved**: `take_note_activity` via `mem::take`.
6. Untested paths — **resolved and DRYed**: one `apply_drum_pad_flash` serves clip activity, note repeat, and live input, with a single test driving all three paths.
7. Naming — **resolved**: `drum_pad_flash_until` / `trigger_drum_pad_flash` / `drum_pad_is_flashing`.

Notes 9 (retain capacity) and 10 (140 ms hold vs fast rolls) stand as informational.

PR #211, opened 2026-08-26
fix: preserve workspace above detail panel

alexanderwanyoike:
## Summary
- cap the detail panel before it can crush Perform or Arrange
- reserve enough height for the app chrome and a usable workspace
- add a regression test for the reported 1032 px window layout

## Test plan
- `cargo test -p vibez-ui detail_panel`
- `cargo test -p vibez-ui instrument_pads_fire_the_visible_drum_rack_bank`

PR #217, opened 2026-08-26
fix: cap Clip and Device panel expansion

alexanderwanyoike:
## What changed

- cap the Clip and Device detail panel at 35% of the current window height
- match the maximum panel composition from the dogfood reference screenshot
- keep the existing absolute workspace reserve for short windows
- cover audio, MIDI and oversized saved panel preferences with responsive tests

## Verification


alexanderwanyoike:

One constant, properly reasoned: the comment ties 35% to the dogfood reference composition (a 1,920x1,040 work area leaving ~364 px), at which height the MIDI editor's 360 px preference exactly fits — clearly the number's design intent. Tests updated coherently, including a correct switch to a float tolerance for the non-representable expectation.

## Nits

1. **note** — at 900 px-class windows the MIDI detail editor now gets 315 px against its 360 px preference (previously 320), slightly deepening the yield policy #211 made explicit and truthful. Consistent with the documented trade; flagging only so the number is chosen, not accidental.

PR #221, opened 2026-09-15
fix: preserve insert effect tails during export

alexanderwanyoike:
## Background

Offline rendering skipped a track's insert effects whenever its audio source or instrument produced silence. Live playback continued processing those effects. This cut off echo and reverb tails during export and left delay history paused until the next source hit, changing subsequent repeats.

## Tasks

- [x] Process track inserts through silent source blocks during offline rendering, matching live playback.
- [x] Add live-versus-export regression coverage for sampler notes and audio clips separated by silence.

## Tested

- Both new regression tests failed before the fix at the first echo block and pass after it. They compare every rendered sample, including gaps and subsequent hits.
- All 244 engine tests passed with `yarn exec -- cargo test -p vibez-engine --locked`.
- A 46.08-second vocal reproduction using the actual sample, MIDI notes, EQ and Tape Echo 2 CLAP settings now matches live playback sample-for-sample. The vocal track uses no sends.
- Formatting and `git diff --check` passed.


alexanderwanyoike:
## mini-code-review: merge as-is

The fix is correct. Both source paths zero the mix buffer before returning (`PlaybackSource::render_audio` fills with 0.0, `render_instrument` zeroes before its early returns), so running inserts on a "nothing produced" block processes true silence, never stale data from the previous block. The offline loop now matches `engine_render.rs:333`, and sends carry effect tails to buses the same way live does. Both regression tests pass locally.

### Bugs
None.

### Design/DRY
- **note** `crates/vibez-engine/src/render.rs:412` The `if track.instrument.is_some() { ... } else { ... };` now evaluates to a `bool` that is silently discarded by the trailing semicolon. It compiles, but a reader has to notice the `;` to see the result is intentionally dropped. Binding it (`let _rendered = ...`) or dropping the expression form makes the intent explicit.
- **note** `crates/vibez-engine/src/render.rs:1533` The test helper builds two unrelated worlds (sampler + note clip vs. audio track + two clips) inside one 200-line function on a `kind` branch; the shared tail is only the echo plugin, transport, and comparison. Two small setup helpers feeding one comparison function would read easier and make a third track kind cheap to add.

### Nits
- **note** `crates/vibez-engine/src/render.rs:1673` The live side relies on `SetArrangementRecording(true)` purely to keep the transport from stopping at the arrangement end. That couples the test to an unrelated feature's side effect; if recording semantics change (for example, capturing input or muting clip playback) this test will fail for a confusing reason. A short comment on why recording is the chosen lever, or an explicit arrangement-length command if one exists, would protect it.
- **note** `crates/vibez-engine/src/render.rs:430` The new comment restates the same rationale as the one at `engine_render.rs:333`. Fine as-is, but per the "one why, one location" rule a pointer to the live path would be enough.

